Issue #2, opened 2016-04-07
Refine unwrap


icanzilb:
`unwrap` was included while we were drafting the repo structure as an example - I got to:
1) add more tests - test for values, edge cases
2) have another look at the code 
3) refine the playground example

I think it's important for this one to is the way we want the code to look lile since this is the first piece of code and others who PR would look at it for guidance how to structure their stuff (maybe?)

sergdort:
Hi, guys correct me if I wrong :)

But I think it would be more understandable for [Optional](https://github.com/RxSwiftCommunity/RxSwift-Ext/blob/master/Source/unwrap.swift#L37) `isEmpty` to be 
```swift
    public func isEmpty() -> Bool {
        return self == nil
    }

```
rather then 
``` swift 
    public func isEmpty() -> Bool {
        return !(flatMap({_ in true})?.boolValue == true)
    }
```

icanzilb:
@sergdort have you tried if the code you're proposing works?

sergdort:
@icanzilb the tests are passing on my machine :)

icanzilb:
ok I'll give it a try 👍

icanzilb:
@sergdort thanks for the suggestion - I made the update to the code already 👍

PR #11, opened 2016-04-12
unwrap tests, examples, docs

icanzilb:
as per https://github.com/RxSwiftCommunity/RxSwift-Ext/issues/2


icanzilb:
@fpillet - few small imporvements over unwrap code and docs as per #2. Could you have a look and leave an okay/not okay comment here before I merge, just a sanity check?

fpillet:
Yes I'm looking at your commits right now. Looks good to me. Let's merge it!

fpillet:
BTW you may want to drop the mt-unwrap branch now, it's been made public and pushed to the repo

Issue #21, opened 2016-04-26
Distinct operator


sergdort:
### Name and description 
`distinct()`

[suppress duplicate items emitted by an Observable](http://reactivex.io/documentation/operators/distinct.html)
### Motivation for inclusion

We have some variants of `distinctUntilChanged` operator in the main repo, but do not have `distinct`, which is implemented for the most other languages which supports Rx.

### Proposed implementation

```swift
extension Observable where Element: Hashable {
   
   func distinct() -> Observable<Element> {
      var set = Set<Element>()
      return flatMap { element -> Observable<Element> in
         objc_sync_enter(self); defer {objc_sync_exit(self)}
         if set.contains(element) {
            return Observable<Element>.empty()
         } else {
            set.insert(element)
            return Observable<Element>.just(element)
         }
      }
   }
}
```

icanzilb:
I'm just thinking aloud here - say if you're using this operator on an Observable<UIImage> wouldn't that consume a lot of memory very quickly? Since the element is a `Hashable` anyway (in your proposed code) isn't it better to store the hashes instead of the elements themselves?

sergdort:
@icanzilb, I see your point.

But, I wonder if could be a case where `Observable` emits elements which  have a different type but they are related ? 
Kind of :
```
class A: Hashable {
}
class B: A {
}

let subject = PublishSubject<A>()

let distinct = subject.asObservable().distinct()

.....

subject.onNext(a)
subject.onNext(b) //so hash of  `b` is the same as hash of `a` but they are different
```

So in this case b would not be edited.

Or I missing something :)

sergdort:
But, yeah potential memory problems is definitely a problem :(

sergdort:
But if we talking about classes. Isn't `Set` will just hold strong reference to the object?

icanzilb:
In your code you're adding the parameter to `Hashable` not to classes. And even if we limit the operator to classes I think this case will be even worse because adding the object to the set will eventually retain it, while it otherwise might have been released (in case nothing else retained the object of course). I think this

icanzilb:
I can see this operator being totally useful, I'm just thinking of a way to have it without a) side effects (e.g. retaining the objects and b) optimal memory footprint (e.g. when the Observable is emitting structs)

sergdort:
@icanzilb  I totally agree with you 

The only concern that I have is in this test case:

```swift
class A: Hashable, CustomStringConvertible {
[...]

PR #23, opened 2016-05-04
Added distinct operator and tests to cover its logic

sergdort:
#21


fpillet:
Fantastic, thanks serg! 

Just one comment, I personally tend to stay away from objc_sync_enter / objc_sync_exit. I didn't even know they worked with non-ObjC objects!

sergdort:
Hi, @fpillet, do you think I would be better to use NSLock instead?

fpillet:
It would be cleaner I think, maybe ask @kzaher what he thinks about it?

kzaherRH:
You don't need locks here

```
    public func distinct() -> Observable<Element> {
 +        var cache = [Element]()
 +        return flatMap { element -> Observable<Element> in
 +            objc_sync_enter(self); defer {objc_sync_exit(self)}
 +            if cache.contains(element) {
 +                return Observable<Element>.empty()
 +            } else {
 +                cache.append(element)
 +                return Observable<Element>.just(element)
 +            }
 +        }
 +    }
```

There is no concurrency since this is a single sequence.

Issue #24, opened 2016-05-04
CI integration setup


sergdort:
Hi, guys 

Since we have some working operators already. 
What do you think about CI ?

icanzilb:
I ❤️ it :) And we also need Danger to take care of first line support (at the very least)

fpillet:
I'm all for it! We have proper tests from the get go, so it totally makes sense

PR #26, opened 2016-05-05
Added .travis.yml

sergdort:
#24 

Hi, guys!

I'm not an expert in Travis, please correct me if I'm wrong :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+# references:
+# * http://www.objc.io/issue-6/travis-ci.html
+# * https://github.com/supermarin/xcpretty#usage
+
+language: objective-c
+osx_image: xcode7.3
+xcode_sdk: iphonesimulator9.3
+before_install:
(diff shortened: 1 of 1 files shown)


fpillet:
I looked at Ash's travis file for Action in https://github.com/RxSwiftCommunity/Action/blob/master/.travis.yml and noticed that you're using `xctool` which is not standard in Xcode (it's a Facebook tool), I'm not sure whether it works with Travis.

Were you able to test this?

sergdort:
Hi, @fpillet 

Yes, I was able to run it on my local machine, because I have `xctool` installed :)

According to this [post](https://www.objc.io/issues/6-build-tools/travis-ci/). Travis comes with `xctool` pre-installed

But if you think it would be better to use standard tool, I don't mind :)

fpillet:
Let's give it a go and see how it goes. How do we see the build results?

PR #35, opened 2016-05-13
Added `distinct` version with `predicate`

sergdort:


delixion:
To satisfy the meaning of 'distinct', I think
`if try cache.contains(predicate)`
should contain `element`, but it does not.
It would be more appropriate to use `comparer` that compares two Elements rather than `predicate` that does not use `element` of .next event.

@sergdort , what do you think?
Do I misunderstand the meaning of 'distinct'?

PR #36, opened 2016-05-15
[WIP] Mt cp1.0.0

icanzilb:
* make playground work with CP 1.0.0
* add back test target for the demo project


icanzilb:
ok, playground is working again with a single scheme + CP 1.0.0 thanks to Boris B from CocoaPods, changes are in this PR

test target - we don't need a separate one anymore, Cmd+U is enough to build + run tests

PR #46, opened 2016-07-27
Add documentation for retry method

reloni:
Hello.
I added playground page for retry method. 
And readme.md now contains section with explanation, how it works.


fpillet:
Thank you very much!

PR #49, opened 2016-08-05
Repeat on complete

icanzilb:
@fpillet - coded this one in the way you suggested - pr includes playground page + tests (mostly copied from the retry op)

since `repeat` is a reserved word in swift I implemented it as `repeatWithBehavior`, what do you think about that? Not coherent with the `retry` operator but it doesn't require backticks to use ...


fpillet:
Thanks, Marin! I agree that `repeatWithBehavior` is not ideal -- I'm merging your PR and we can review our options before release, maybe rename both operators for consistency?

PR #51, opened 2016-08-19
Fixed building with Carthage

tcurdt:
What's still left to do is to run the test and wrap the project with cocoapod so it works with both.


fpillet:
Ouch just noticing this PR now. Guess I've been too much head in code. Right now it doesn't pass CI, could you look into it? Thanks!

PR #54, opened 2016-09-19
Updated to support Swift 3.

MadsBogeskov:


MadsBogeskov:
afaik the reason Travis fails is that this should be compiled with the new Swift compiler.

Demo/RxSwiftExtDemo/RxSwiftExtPlayground.playground/SupportCode.remap
```diff
@@ -0,0 +1,53 @@
+[
+ {
+  "file": "/Users/dkmadboe/Developer/Github/RxSwiftExt2/Demo/RxSwiftExtDemo/RxSwiftExtPlayground.playground/Sources/SupportCode.swift",
```
fpillet: There shouldn't be absolute paths in your playground -- unless this file should not be committed?

fpillet:
I'm not worried about the CI issues at this stage - I just need to move current to a swift-2 branch before merging your commit (will give you write access to the repo as well)

Sorry about the delay responding - I've been quite busy this week

Demo/RxSwiftExtDemo/RxSwiftExtPlayground.playground/SupportCode.remap
MadsBogeskov: Good catch. Consider it gone in 3... 2...

MadsBogeskov:
@fpillet no problem. We have just migrated to Swift 3 at the office, so it would be great to have this in, instead of using my own fork :)

fpillet:
I have merged this PR in the swift3 branch as of commit ccbc4e8c3281dae593dbd1405b542afbaa4044b5

Thanks! Let me know if referencing this branch from your code works for you. And thanks a lot for taking care of the swift3 transition!

PR #61, opened 2016-11-06
Fix Travis build

reloni:
Hello.
I made some changes in project in order to fix Travis-CI builds. Main thing that I did: I changed SWIFT_VERSION from 3.0.1 to 3.0, because Travis builds project with XCode 8.0 and it seems XCode doesn't understand newer Swift version:)
I also changed Travis config, now it uses xcodebuild to directly build two Rx schemes (RxSwift and RxTests), this fix significantly reduces Travis build time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ DerivedData
 .bundle
 
 # Carthage
-
+Carthage
 !Carthage/**
 
 # Cocoapods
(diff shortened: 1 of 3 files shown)


fpillet:
Fantastic, thanks for contributing this! I looks like you also fixed the issue where when using RxSwiftExt from Carthage, it doesn't embed the RxSwift framework in the RxSwiftExt.framework anymore, right?

reloni:
@fpillet 
Mmm, if I understand you correctly: yes, RxSwift not embedded into RxSwiftExt framework now.

PR #63, opened 2016-11-24
Change iOS deployment target to 8.0 in Xcode project

mr-v:
Hi,
I've changed project's deployment target to match one from podspec. This allows to integrate library with Carthage into pre-iOS 9.3 app.


fpillet:
Thanks!

PR #68, opened 2017-01-06
Add materialize/dematerialize operators

acchou:
Includes the operators, tests, playground page, and Readme doc.
I noticed that the instructions for Carthage don't have --no-use-binaries, so I added that to the Contributing doc. But I didn't make the change to the playground files. Might want to consider that as well.


fpillet:
Thanks a lot!

PR #71, opened 2017-01-13
Update RxSwift version in cartfile

reloni:
This allows Travis-CI to run unit tests with RxSwift 3.1.0


fpillet:
Yeah it's you're right, it's about time we just keep in sync with RxSwift. Thanks!

PR #72, opened 2017-01-15
Refactoring

adamnemecek:


fpillet:
@adamnemecek Thank you for submitting this.

As is, I can't merge this PR. But read on!

- First thing is this project uses tabs by default, we're not going to change to spaces. I believe in every developer's right to chose her tab width, whereas with spaces you're locked in whatever choice the original dev made. But let's not enter a religion war over this.
- Some interesting changes in removing `Observable<T>` here and there. I agree that these could be removed. 👍  **Would you mind submitting a separate PR for this?**
- I'm not sure about the change from a `for` loop to a range + `.contains` in `cascade.swift`. It creates an object on the heap plus a closure (capturing a property), overall it's probably a negative gain in speed for harder-to-read code. Sometimes, less isn't more :)
- Good one with change from `contains` to `insert` in cascade, I like it! 👍  **Would you mind submitting a separate PR for this?**

Thank you very much for helping improve RxSwiftExt! I know it can be frustrating when you spend time on something and it doesn't get merged, but only some of your changes won't make it.

Also interested in other contributors opinion about the various changes. @icanzilb ? @reloni? @sergdort?

reloni:
I like this changes.
For me sometimes a little verbose code helps to understand what happening, so I think that `Observable<T>.` is  ok, but I agree that we can and should remove it.

PR #74, opened 2017-02-08
Move from swift 3.0.1 to 3.0 for better compatibility

joihelgi:
There are issues with Travis CI, carthage (w/o -no-use-binaries) and other libs at swift 3.0. I do not see the reason for the move to swift 3.0.1 for libraries until there is full support from CIs and libraries like RxSwift make the move.


joihelgi:
@tcurdt thanks

PR #84, opened 2017-04-14
Move materialized+elements to Source directory

TheAdamBorek:
The file with `elements()` & `errors()` was in main directory 😅. As a result, these operators are not visible in newest version of the library

Before:
zrzut ekranu 2017-04-14 13 47 21

After:
zrzut ekranu 2017-04-14 13 51 51


fpillet:
Thanks Adam

PR #90, opened 2017-05-31
CustomOperator and FilterMap

jeremiegirault:
I had to implement `filterMap` for my own use and I had the idea to make it a bit more generic using a kind of `flatMap` using a `CustomOperator` type.

I'm providing both to the community.


fpillet:
Great, thanks!

PR #97, opened 2017-07-14
Added tvOS support

MadsBogeskov:


fpillet:
Thanks! As it is, I can't merge this PR. Tests fail for me.

In order of what I noticed in the targets:

- RxSwiftExt-tvOS doesn't link with the RxSwift framework
- RxSwiftExtTests-tvOS doesn't link with the RxSwift and RxTests frameworks
- RxSwiftExtTests-tvOS doesn't include the `Copy Carthage Framework` phase at the end which is required to copy the frameworks to the proper location for the tests to run.

MadsBogeskov:
Well that was embarrassing. I have now successfully run the tests here, and pushed the changes. Hoping for Travis to be happy as well :)

fpillet:
Thanks both ! Merging

PR #98, opened 2017-07-21
[Proposal] Deprecates mapTo in favor of map(to:).

noppefoxwolf:
# Proposal

In `RxSwift`, `bindTo` was become `bind(to:)` at 3.4.0.
In the same way, I think that `mapTo` should be `map(to:)`.


fpillet:
Good thinking! Would you mind amending your PR with an update to the Playground example page for `mapTo` (index will need an update as well) and the README ? Thanks!

noppefoxwolf:
@fpillet 
Thank you for your response.
Fix Playground and Readme.
The file name of Playground remains `mapTo.xcplaygroundpage`, but is there a problem?
This is because it is hard to distinguish from the original `map`.

fpillet:
Nearly there! Link to the `map(to:)` page from playground Index page is still broken. Other than this, changes are 👍

fpillet:
That was fast. Thanks!

noppefoxwolf:
@fpillet Thank you for your polite response.

PR #110, opened 2017-10-18
RxSwiftExt 3.0.0

fpillet:
Complete changes for 3.0.0


freak4pc:
@fpillet Just wondering, do you think it might be better to name as 4.0.0 to be on par with RxSwift, or does that seem wrong to you?

fpillet:
@freak4pc I don't think we need to bump in lockstep -- as long as the Readme indicates what version of RxSwift we relate to. If you look at RxDataSources, it just bumped to 3.0.0 not 4.0.0.

PR #111, opened 2017-10-18
Combine observables using the .and(...) operator.

jdisho:


Source/RxSwift/and.swift
```diff
@@ -0,0 +1,22 @@
+extension Observable where E == Bool {
+
+   func and(_ right: Observable<Bool>) -> Observable<Bool> {
```
ashfurrow: I think these should be `Observable<E>`, right?

joanii: Correct, I will refactor it 👍

fpillet:
Thanks for your contribution @joanii ! Two things about this:

1. We require that any code added to the repository also adds tests to validate it. This may seem trivial but the whole chain of trust relies on these tests to guaranteee that what we expose works as advertised

2. I think that it may be beneficial to add a few other signatures:

`public static func and<C: Collection>(_ collection: C) -> Observable<Bool>
    where C.Iterator.Element: ObservableType, C.Iterator.Element.E == E`
and
`public static func and<O: ObservableType>(_ sources: O ...) -> Observable<Bool>
    where O.E == Bool`

and maybe instance variants as well for completeness.

What do you think?  Sorry for the delay coming back to you.

fpillet:
Oh and also will need to add a Playground page to the test project with examples for `and` :)

Thanks!

joanii:
Thank you for the feedback @fpillet

joanii:
Can you explain shortly the purpose of other signatures that you suggested? 
I am somehow new to the RxSwift community and it feels a bit hard to follow it :) @fpillet

fpillet:
@joanii I see that you added variants that take two, three, four observables to make a combined `and`. I think it would be beneficial to keep only two variants, one that takes a `Collection` of observables, and one that takes a variable number of arguments of the same type. This would cover all possible use cases.

joanii:
You can review it now 👍 @fpillet

PR #114, opened 2017-10-25
Refactor the unwrap code.

joanii:
Making the unwrap function with less lines of code and more readable. 
This function can also be called "skipNil" because it is skipping nil values.


icanzilb:
Skipping nil values is not what this op does. In unwraps optional values

PR #115, opened 2017-10-25
[Proposal] Refactor unwrap

joanii:
Make *unwrap* function concise and easier to follow.


fpillet:
Ah this looks much better! I like this version, nice and concise

PR #116, opened 2017-10-31
Added app extension support

MadsBogeskov:
This PR adds the APPLICATION_EXTENSION_API_ONLY flag to all targets. This is necessary if RxSwiftExt should be used in an app extension, like the top shelf on tvOS. It will basically just warn if any APIs that aren't supported in extensions are used.

The other thing that I added (and I can remove quickly if unwanted) is that the sources folder was pointing to an invalid location on disk called NSObject_Rx.

### DIFF
--- a/RxSwiftExt.xcodeproj/project.pbxproj
+++ b/RxSwiftExt.xcodeproj/project.pbxproj
@@ -339,7 +339,7 @@
 				BF515CDE1F3F36F200492640 /* Tools */,
 			);
 			name = Sources;
-			path = NSObject_Rx;
+			path = Source;
 			sourceTree = "<group>";
 		};
 		18EE7A0F1C47B12F00C7256C = {
@@ -382,8 +382,7 @@
 				5386076E1E6F1C0A000361DE /* mapTo+RxCocoa.swift */,
 				538607EF1E6F589E000361DE /* not+RxCocoa.swift */,
 			);
-			name = RxCocoa;
-			path = Source/RxCocoa;
+			path = RxCocoa;
 			sourceTree = "<group>";
 		};
 		538607701E6F1C38000361DE /* RxCocoa */ = {
@@ -421,7 +420,7 @@
 				98309EB01EDF159500BD07D9 /* filterMap.swift */,
 				BF515CE11F3F371600492640 /* fromAsync.swift */,
 			);
-			name = RxSwift;
+			path = RxSwift;
 			sourceTree = "<group>";
 		};
 		538607BA1E6F362B000361DE /* RxSwift */ = {
@@ -485,8 +484,7 @@
 			children = (
 				BF515CDF1F3F370600492640 /* curry.swift */,
 			);
-			name = Tools;
-			path = Source/Tools;
+			path = Tools;
 			sourceTree = "<group>";
 		};
 		E36BDFB81F38755F008C9D56 /* tvOS */ = {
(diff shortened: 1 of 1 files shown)


freak4pc:
This is a great change !

I'm wondering if there's any way to add to the tvOS target an extension target as well and add it to Travis to make sure this fix is correctly tested. Would you mind adding this in ? @MadsBogeskov

MadsBogeskov:
So the settings just disables certain APIs (like UIApplication.shared) from being called, and so I don't know how you would test that it works?

freak4pc:
I guess the point of this is to make it work on extensions. So “testing” is adding an extension target with some dummy code and make sure Travis compiles it (IMO)

bobgodwinx:
Hi guys I thought this was the job of `post_install` hook ? I mean setting the flag `APPLICATION_EXTENSION_API_ONLY` = `YES` correct me if am wrong.

TeamSpartans:
@bobgodwinx when I use the library in my tvOS top shelf extension it tells me the flag isn't set so I guess it doesn't. But wouldn't you like to have it in the project it self?

@freak4pc but it wont be able to compile with the flag enabled if you have set the flag.

MadsBogeskov:
@bobgodwinx when I use the library in my tvOS top shelf extension it tells me the flag isn't set so I guess it doesn't. But wouldn't you like to have it in the project it self?

@freak4pc but it wont be able to compile with the flag enabled if you have set the flag.

freak4pc:
Uhm not sure how that makes sense. Maybe we should look into other project that support tvOS and see if they set this flag. Could you find a library similar to this? 

I’m wondering if @bobgodwinx is right and maybe this is an app-level configuration usually.

MadsBogeskov:
This is not specific to tvOS. This is a general flag set for app extension frameworks. The flag just tells the compiler that certain APIs cannot be used, and so if you are using them in the framework, it wont be able to compile.
So RxSwiftExt _doesnt_ use any of these APIs, so there isn't a problem at the moment. However I do get a warning in my project about using a framework in an extension that isn't guaranteed to be extension capable, because the flag wasnt set. The problem here is that if someone, at some point, uses an API that isn't approved for extension use, then as far as I understand, you wont get an error before uploading to iTunes Connect.

freak4pc:
I see what you're saying. The thing is - Yes, right now RxSwiftExt doesn't use any of these APIs, but would it be correct to assume that would always be the case? Shouldn't we just expect extension developers to make sure they have a build step that adjusts this flag for libraries ?

bobgodwinx:
The rule of thump here is that developers configure their settings the way it suites.  `APPLICATION_EXTENSION_API_ONLY` is one of this and it makes sense to let the developer do it. Having said that, if everyone is using this `FLAG` we can then make it default and add it to the framework

groue:
@freak4pc I suggest that you set the flag now, and prevent right away RxSwiftExt from using extension-forbidden APIs. As soon as one would sneak in, it would be difficult to remove it. The proper answer to forbidden APIs, I think, is **injection**.

Let's take an example: `UIApplication.shared` is not available to extensions, but `UIApplication` is. This means that extensions are not allowed to fetch the current app by themselves (which may not exist), but they are allowed to use an app that has been given to them. When an extension-compatible library has some nice feature based on `UIApplication.shared`, it thus documents somewhere that an UIApplication instance has to be provided in some `activateNiceFeature(application: UIApplication)` method. The application developer calls this method, and moves on to his next task. The extension developer has no application instance to share with the library, thus understands why the nice feature isn't active, and also moves on to his next task.

freak4pc:
This makes sense to me. I just want to get @fpillet's blessing as this is his mainly his repo :)

fpillet:
@freak4pc Please go ahead. I'm currently heads down in the bowels of iOS networking issues, can't get my mind of this to test this proposal. @groue 's arguments make total sense to me.

freak4pc:
Would still be cool to have a test target on Travis to build an extension (so we constantly confirm we're not breaking). but this is great for now. LGTM! :shipit:

PR #123, opened 2017-12-09
nwise and pairwise operators

### LINKED ISSUE #42: pairwise, nwise operators
### parwise(), nwise(n:)

Transforms a sequence to return every consecutive pair of emissions.
### Motivation for inclusion

One often need to compare a previous value for changes, this would make that simpler than the rather convoluted combination of .scan, .filter, .map needed to achieve this currently.
### Example of use

```
-----0-----1-----2-----3-----4----->
.parwise()
     v     v     v     v     v
---------(0,1)-(1,2)-(2,3)-(3,4)--->

```

vzsg:
This PR resolves #42 using scan/filter/map.
I tried to follow the guidelines and include tests and a playground page.

### DIFF
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -37,6 +37,7 @@
  - [pausableBuffered()](pausableBuffered) operator, pauses emission of elements unless the most recent element from the provided sequence is `true`; buffers elements that arrive while paused; emits buffered elements on resume.
  - [filterMap()](filterMap) operator, filters out some values and maps the rest (replaces `filter` + `map` combo)
  - [Observable.fromAsync()](fromAsync) constructor, translates an async function that returns data through a completionHandler in a function that returns data through an Observable
+ - [nwise(), pairwise()](nwise) operators, which group the items emitted by an Observable into arrays of N or pairs of consecutive items
 */
 
 //: [Next >>](@next)
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -32,12 +32,12 @@
  - [distinct()](distinct) operator, suppress duplicate items emitted by an Observable
  - [retry()](retryWithBehavior) operator, retry on error with various delay / predicate options
  - [repeatWithBehavior()](repeatWithBehavior) operator, repeat on completion with various delay / predicate options
+ - [nwise(), pairwise()](nwise) operators, which group the items emitted by an Observable into arrays of N or pairs of consecutive items
  - [catchErrorJustComplete()](catchErrorJustComplete) ignore any error and complete the sequence instead
  - [pausable()](pausable) operator, pauses emission of elements unless the most recent element from the provided sequence is `true`
  - [pausableBuffered()](pausableBuffered) operator, pauses emission of elements unless the most recent element from the provided sequence is `true`; buffers elements that arrive while paused; emits buffered elements on resume.
  - [filterMap()](filterMap) operator, filters out some values and maps the rest (replaces `filter` + `map` combo)
  - [Observable.fromAsync()](fromAsync) constructor, translates an async function that returns data through a completionHandler in a function that returns data through an Observable
- - [nwise(), pairwise()](nwise) operators, which group the items emitted by an Observable into arrays of N or pairs of consecutive items
 */
 
 //: [Next >>](@next)
--- a/Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
@@ -14,10 +14,10 @@ import RxSwiftExt
 /*:
  ## nwise
 
- The `nwise` operator groups elements emitted by an Observable into arrays, where each array consists of the last N consecutive items; similar to a sliding window.
+ The `nwise` operator groups elements emitted by an Observable into arrays, where each array consists of N consecutive items; similar to a sliding window.
  */
 
-example("Grouping numbers into threes") {
+example("Grouping numbers into triples") {
     let input = [1, 2, 3, 4, 5, 6]
 
     print("Input:", input)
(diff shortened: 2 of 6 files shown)


Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,53 @@
...
+        .nwise(3)
+        .subscribe(onNext: { result in
+            print(result)
+        })
+}
+
+/*:
+ ## pairwise
+
+ The `pairwise` operator is a special case of `nwise` with the size of 2, which emits the last and previous items in tuples instead of arrays.
```
freak4pc: > which emits the last and previous items in tuples instead of arrays.

should this be "current and previous items" ? (or current and next, depends on how you look?)

Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
```diff
@@ -37,6 +37,7 @@
  - [pausableBuffered()](pausableBuffered) operator, pauses emission of elements unless the most recent element from the provided sequence is `true`; buffers elements that arrive while paused; emits buffered elements on resume.
  - [filterMap()](filterMap) operator, filters out some values and maps the rest (replaces `filter` + `map` combo)
  - [Observable.fromAsync()](fromAsync) constructor, translates an async function that returns data through a completionHandler in a function that returns data through an Observable
+ - [nwise(), pairwise()](nwise) operators, which group the items emitted by an Observable into arrays of N or pairs of consecutive items
```
freak4pc: I think this logically fits better under "repeatWithBehavior" (would defer to @fpillet for confirmation)

Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+//: [Previous](@previous)
+import RxSwift
+import RxSwiftExt
+
+/*:
+ ## nwise
+
+ The `nwise` operator groups elements emitted by an Observable into arrays, where each array consists of the last N consecutive items; similar to a sliding window.
```
freak4pc: The term "last" is confusing here. can you explain what you meant? 
Seems to me like its "each array consists of N consecutive items"

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+     - emits an array for every element after that;
+     - and forwards any error or completed event.
+
+     For example:
+
+          --(1)--(2)--(3)-------(4)-------(5)------->
+           |
+           | nwise(3)
+           v
+          ------------([1,2,3])-([2,3,4])-([3,4,5])->
```
freak4pc: 👏 👏 👏 👏 👏  Fan of ASCII diagrams !

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+
+     Caveat: n = 0 causes the operator to collapse into: `.map { _ in [] }`.
+
+     - parameter n: size of the groups
+    */
+    public func nwise(_ n: UInt) -> Observable<[E]> {
+        return self
```
freak4pc: I understand the use for Uint (non-negative), but I think having Uint and then performing an Int conversion is a bit unneeded and is a bit of boilerplate.

I would just have make it an Int and assert the input.

Two options that come to mind
1. `assert(n > 0, "n must be positive")`, or guard against it and emit an error (might even make sense to assert  `> 1`)
2. use abs(n), but this might confuse a developer.

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+            .scan([]) { acc, item in
+                // UInt->Int conversion is used to enforce a non-negative parameter at build time.
+                // Supplying a negative number to nwise would cause a fatal error at runtime.
+                Array((acc + [item]).suffix(Int(n)))
+            }
+            .filter { $0.count == n }
+    }
+
+    /**
+     Groups the elements of the source observable into tuples of the last and the previous elements.
```
freak4pc: same question here - should this be current and previous ?

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+
+     The resulting observable:
+     - does not emit anything until the source emits at least 2 elements;
```
freak4pc: "the source observable" or "the source sequence"

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+     - emits a tuple for every element after that;
+     - and forwards any error or completed event.
```
freak4pc: - forwards any error or completed events. (no and, and event is plural)

Source/RxSwift/nwise.swift
```diff
@@ -0,0 +1,62 @@
...
+            }
+            .filter { $0.count == n }
+    }
+
+    /**
+     Groups the elements of the source observable into tuples of the last and the previous elements.
+
+     The resulting observable:
+     - does not emit anything until the source emits at least 2 elements;
+     - emits a tuple for every element after that;
```
freak4pc: makes sense to be "... for every element after that; concluding of the current and previous elements" (or something similar)

vzsg:
Thanks for the review! I'll get to fixing the issues right away.

Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
vzsg: To be honest, I haven't figured out the best way to phrase something so visually trivial. This is why I have three similar, yet different versions in all places...

Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
vzsg: Definitely "current and previous" out of the three.

Playground/RxSwiftExtPlayground.playground/Pages/nwise.xcplaygroundpage/Contents.swift
freak4pc: What about just simplifying it as:
> The `nwise` operator groups elements emitted by an Observable into arrays of `n` items each

Source/RxSwift/nwise.swift
vzsg: I'll change this to an `assert` then.

Source/RxSwift/nwise.swift
vzsg: Thanks, me too :) And I was pleasantly surprised how well it looks in Xcode's quick help bubbles.

fpillet:
Fantastic, thanks both for the hard work!

PR #134, opened 2018-01-19
OfType Operator

glwithu06:
- Implement `ofType` operator using `filter` and `map` operator.
- Add `ofType` extension methods on `ObservableType`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Changelog
 
 - added `mapAt(keyPath:)` operator
 - added `zip(with:)` operator
-- added `ofType()` operator
+- added `ofType(_:)` operator
 
 3.1.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 - added `mapAt(keyPath:)` operator
 - added `zip(with:)` operator
+- added `ofType()` operator
 
 3.1.0
 -----
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -39,6 +39,7 @@
  - [pausableBuffered()](pausableBuffered) operator, pauses emission of elements unless the most recent element from the provided sequence is `true`; buffers elements that arrive while paused; emits buffered elements on resume.
  - [filterMap()](filterMap) operator, filters out some values and maps the rest (replaces `filter` + `map` combo)
  - [Observable.fromAsync()](fromAsync) constructor, translates an async function that returns data through a completionHandler in a function that returns data through an Observable
+ - [ofType()](ofType) operator, filters the elements of an observable sequence, if that is an instance of the supplied type.
 */
 
 //: [Next >>](@next)
--- a/Playground/RxSwiftExtPlayground.playground/Pages/ofType.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/ofType.xcplaygroundpage/Contents.swift
@@ -1,0 +1,33 @@
+/*:
+ > # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+ 
+1. Make sure you have [Carthage](https://github.com/Carthage/Carthage) installed
+1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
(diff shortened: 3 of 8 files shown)


glwithu06:
Actually, I have implemented this operator in `RxSwift`.
And @kzaher said, `RxSwiftExt` is a better place to put this functionality.
Here is my PR on `RxSwift`. (https://github.com/ReactiveX/RxSwift/pull/1529)

freak4pc:
Didn't have time to look into this one yet, sorry about that. Hope to find some time later :) 
In general though, I would name it `filterType` or `takeType` and not `ofType`, since the former is in-line for discovery of filtering operators usually.

freak4pc:
@glwithu06 Let me know if you still want to add this one. 

If we call it `filterType` or `takeType` ( I think the latter makes more sense ), I'm good with adding it :)

jegnux:
Oops, I didn't see this PR and open a similar one. But let's keep the discussion here.
In my PR I called they operator `filterMap(as:)`
Here is why I think it's the best option.

- `filterType`, `takeType` and `ofType` doesn't reflect imho that the result is actually map
- `filterMap` does reflect this information
- `as` is common to express casting in swift world
- `filterMap(as:)` matches the name style of `map(to:)`

WDYT?

freak4pc:
I'm not super religious to this, I get your point here, but I'm not entirely sure. 
I consider the fact its a map to be an implementation detail, while the main idea of it is it would not pass items that are not of a certain type. 

takeType wouldn't fit here since take has the "termination" kind of behavior, so I think it stands between 

`filterType(Int.self)` vs. `filterMap(as: Int.self)` 

@fpillet if you have a minute would love your call here :) I think we need a third set of eyes

jegnux:
> takeType wouldn't fit here since take has the "termination" kind of behavior,

I'm not sure to understand.`take()` or `takeWhile()` indeed change how the original sequence is completed. But that's not the case of this new operator. This new operator completes when the original sequence completes.

jegnux:
Hum... mixed feelings coming... It seems that other Rx implementations of this operator use `ofType` as suggested by @glwithu06 

http://reactivex.io/documentation/operators/filter.html

What's the policy of RxSwiftExt? Do we want to match as much as possible other Reactive X implementations?

freak4pc:
Blah, I don't like that name :( Not entirely sure on policy here but we can go with ofType if its already common.

jegnux:
Personally I won’t mind to have a different name than other implementation. 
We could also implement both (ofType and filterMap(as:))

We already have ignoreErrors() in RxSwiftExt as an alias to retry() and RxSwift has debounce() and throttle() that do the same thing.

bobgodwinx:
😂 Naming convention the most difficult aspect of software engineering 😂  now don't shoot me alive but after reading the entire thread this is my suggestion IMHO I would name it as follows.
- ` public func whereType<T>(_ type: T.Type) -> Observable<T> `
Somehow it sounds cool 😎 or whatever 👍

freak4pc:
I don't like that - it doesn't belong to any family of operators from Rx at all. 

I think I agree with @jegnux - to my distaste (🤣), let's make it ofType to conform to RxJava so users of it can find their way easily.

jegnux:
@bobgodwinx I agree with @freak4pc, `where` doesn't have any precedence in Rx world.

>I think I agree with @jegnux
@freak4pc ok to have both `filterMap(as:)` and `ofType()` ?
I strongly believe `filterMap(as:)` is a good choice. It's basically the condensed version of what it does ( `.filter { $0 is T }.map { $0 as! T }`). #lobbying

bobgodwinx:
@freak4pc @jegnux I totally agree that `whereType()` is `SQL` thing 😂 , `ofType()`  on the other hand is bit grrrrrrr. Yes `filterMap(as:)` might fit. I would go with that. 👍

freak4pc:
Let’s go for ofType with an alias with filterMap(as:) and cover all basis :)

jegnux:
A consensus, great !
Now the question is, do we go with my PR (#139) or this one?

freak4pc:
@jegnux since you seem to be available, and no response from OP here, let’s wrap up your PR

glwithu06:
Oops! Sorry for very late response.
I was away for a while.
Anyway we decided to go for 'ofType'. right?

freak4pc:
Hey @glwithu06 - Could you please 

1. Rebase your changes so the merge commit is gone (If you don't know how I can take care of it later worst case)

2. Fix the CI so build passes successfully :) 

Thanks !

fpillet:
I'm fine with `ofType(_:)`, I don't have a better idea to name this operator !

freak4pc:
Made a tiny adjustment to the CHANGELOG (method signature was wrong). 
This looks great ! Thanks for your work @glwithu06, and welcome to RxSwiftCommunity!

PR #135, opened 2018-01-21
add a pausable(afterCount:) operator

twittemb:
The pause is available only after a certain count of events. Before the number of emitted events reaches that count the Pauser will not be taken care of. When the Pauser is active, elements are ignored unless the second sequence has most recently emitted `true`, this becomes the same behavior as the normal pausable operator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog
 - added `apply` for `Completable` and `Maybe` 
 - added `mapTo` for `Single` and `Maybe`
 - added SPM support
+- added `pausable(afterCount:)` operator
 
 5.0.0
 -----
--- a/Playground/RxSwiftExtPlayground.playground/Pages/pausableAfterCount.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/pausableAfterCount.xcplaygroundpage/Contents.swift
@@ -1,0 +1,40 @@
+/*:
+ > # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+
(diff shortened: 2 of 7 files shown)


twittemb:
Hi 

I don't know if that kind of operator could be useful (it is on my project RxFlow :-)) ... 
Anyway I have some issue writing the unit tests. Any help would be appreciated 👍

twittemb:
Hi @freak4pc ... Do you think this PR has an interest ?

Thx.

freak4pc:
@twittemb I just want to confirm, this operator would pause until a certain number of emissions occur on a second Observable ? Didn't really have time to review this and the other guys have been kind of busy lately it seems :)

twittemb:
Hi @freak4pc 
What it does is: the pauser observable is ignored until the first observable has trigger the "aftetCount" number of events.

It is useful to be sure that the first observable will trigger something before being pausable.

freak4pc:
I really want to help out here but I don't have enough experience with the original pausable operator at all :( If someone else could poke at this that'd be great, otherwise I'll see what I can do next week.

twittemb:
Hi here,

just a little UP on this PR 😀

It is OK to drop it if there is no value to bring to this repo.

fpillet:
@twittemb don't close it! We're just being a little overwhelmed :)

freak4pc:
The main issue with the specific one is that I just have no understanding of how it works and what it does so I prefer not to review it - if there's someone who does :)

MortyMerr:
Hey @twittemb I added tests on your operator and extend generic type name. Also I updated readme to include a new operator. Please approve that you r ok with these changes.

MortyMerr:
Hey @freak4pc could u please check this? I resolved conflicts. I am not sure about motivation to add this, so both close/merge is ok for me

PR #136, opened 2018-02-09
Add mapAt(keyPath: KeyPath)

michaelavila:
The goal of this PR is to introduce an operator that allows you to map from some value to some other value, given a key path. An example of the usage follows:

```
struct Person {
    let name: String
}

let people: [Person] = [
    Person(name: "Bart"),
    Person(name: "Lisa"),
    Person(name: "Maggie")
]

Observable.from(people)
    .mapAt(\.name)
    .subscribe(onNext: { print($0) })

// Output:
// Bart
// Lisa
// Maggie
```

My organization is probably incorrect. Let me know if there's anything I can change.

### DIFF
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -26,6 +26,7 @@
  - [ignore(Any...)](ignore) operator, filters out any of the elements passed in parameters
  - [Observable.once(Element)](once) contructor, creates a sequence that delivers an element *once* to the first subscriber then completes. The same sequence will complete immediately without delivering any element to all further subscribers.
  - [mapTo(Any)](mapTo) operator, takes a sequence of elements and returns a sequence of the same constant provided as a parameter
+ - [mapTo(KeyPath)](mapToKeyPath) operator, takes a sequence of elements and returns a sequence of those elements mapped to their values at the given key path
  - [not()](not) operator, applies a the boolean not (!) to a `Bool`
  - [and()](and) operator, combines Bool values from one or more sequences and emits a single `Bool` value.
  - [Observable.cascade([Observable])](cascade) contructor, takes a ordinary sequence (i.e. Array) of observables and cascades through them
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -25,8 +25,8 @@
  - [unwrap()](unwrap) operator, takes a sequence of optional elements and returns a sequence of non-optional elements, filtering out any `nil` values
  - [ignore(Any...)](ignore) operator, filters out any of the elements passed in parameters
  - [Observable.once(Element)](once) contructor, creates a sequence that delivers an element *once* to the first subscriber then completes. The same sequence will complete immediately without delivering any element to all further subscribers.
+ - [mapAt(KeyPath)](mapAt) operator transforms a sequence of elements where each element is mapped to its value at the provided key path
  - [mapTo(Any)](mapTo) operator, takes a sequence of elements and returns a sequence of the same constant provided as a parameter
- - [mapAt(KeyPath)](mapAt) operator, takes a sequence of elements and returns a sequence of those elements mapped to their values at the given key path
  - [not()](not) operator, applies a the boolean not (!) to a `Bool`
  - [and()](and) operator, combines Bool values from one or more sequences and emits a single `Bool` value.
  - [Observable.cascade([Observable])](cascade) contructor, takes a ordinary sequence (i.e. Array) of observables and cascades through them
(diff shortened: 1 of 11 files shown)


freak4pc:
I'll try to take a look at this later, but at first look aren't you mapping *from* and not *to* ? `mapFrom(keyPath)` seems more appropriate, or if an overload works here `map(from keyPath: KeyPath)`

michaelavila:
@freak4pc maybe, I was thinking of this as mapping to the value at the key path. Are you thinking of it as creating a mapping from a key path? That makes sense. I'm not too concerned about this, so I'll follow your lead if you have a preference.

freak4pc:
Oh I meant to open a discussion, not necessarily immediately change :) 

I think the "source" your mapping from is an observable, and the destination is also an observable, nota a keypath. 

So I think my top 3 would be:

- `mapFrom(keyPath: KeyPath)`
- `mapKeyPath(_ keyPath: KeyPath)`
- `map(from keyPath: KeyPath)` - i think this might be bad since we have some bad experiences with overloads of map

@fpillet would love your thoughts

I'll review this code/docs wise a bit later :) 

Thank you @michaelavila !

michaelavila:
@freak4pc of those options I like `mapFrom(keyPath: KeyPath)`. Let me know if I should just make that change.

michaelavila:
@freak4pc can you elaborate more and why you think `from` makes more sense? Admittedly, I still don't get it.

bobgodwinx:
@michaelavila I think @freak4pc was referring to the problem of [map](https://github.com/RxSwiftCommunity/RxSwiftExt/commit/1a7cf0f73dbac48713501ddeee2ccad871ddfa37). IMHO `map(from keyPath: KeyPath)` makes more sense. But if it shows errors use `mapFrom(keyPath: KeyPath)`

fpillet:
Since we already have  `mapTo` I think what makes more sense in the context of RxSwiftExt is to use `mapFrom`

freak4pc:
keyPath is an index for all sorts
When you map to something - you expect to get what you’re mapping to, but you’re not returning a key path here.

mapAt(keyPath:)

Also makes a lot of sense to me since at is constantly used in Swift for indices @michaelavila

michaelavila:
@bobgodwinx , @fpillet, @freak4pc OK. I'm buying this. I'll make the necessary changes and push. Thanks for the feedback.

michaelavila:
Done. I went with `mapAt<R>(_ keyPath: KeyPath<E, R>) -> Observable<R>`

Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
```diff
@@ -26,6 +26,7 @@
  - [ignore(Any...)](ignore) operator, filters out any of the elements passed in parameters
  - [Observable.once(Element)](once) contructor, creates a sequence that delivers an element *once* to the first subscriber then completes. The same sequence will complete immediately without delivering any element to all further subscribers.
  - [mapTo(Any)](mapTo) operator, takes a sequence of elements and returns a sequence of the same constant provided as a parameter
+ - [mapAt(KeyPath)](mapAt) operator, takes a sequence of elements and returns a sequence of those elements mapped to their values at the given key path
```
freak4pc: Can you move this above mapTo (so its alphabetic within the group of maps)

Playground/RxSwiftExtPlayground.playground/Pages/mapAt.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import RxSwift
+import RxSwiftExt
+
+/*:
+ ## mapAt(KeyPath)
+
+ The `mapAt` operator takes a sequence of elements and returns a sequence where each element is mapped to its value at the given key path
+ */
+example("replace any input with the value at key path") {
```
freak4pc: replace any input with its value at the provided keypath *

Readme.md
```diff
@@ -181,6 +181,30 @@ next(Nope.)
 completed
 ```
 
+#### mapAt
+
+Replace every element with its value at the given key path.
```
freak4pc: Its not urgent, so I don't mind merging without change, but I think the term replacing is really confusing - map is a transform method, I would expect something like 

"Transform every element to its value at a given keypath" 

If we change it , then we'll later change this on mapTo as well since I don't think the replace word is appropriate

Tests/RxSwift/MapAtTests.swift
```diff
@@ -0,0 +1,57 @@
...
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import XCTest
+
+import RxSwift
+import RxSwiftExt
+import RxTest
+
+struct Person {
```
freak4pc: You can move this struct inside the MapAtTests class - It's not needed outside of the scope of this file/test

Tests/RxSwift/MapAtTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        observer = scheduler.createObserver(String.self)
+
+        _ = Observable.from(people)
+            .mapAt(\.name)
+            .subscribe(observer)
+
+        scheduler.start()
+    }
+
+    func testReplaceWithResultCount() {
```
freak4pc: What do you think about the naming of the tests? I'm just thinking out loud here if the name is proper to what its testing

testMapAtKeypathCount()
testMapAtKeypathValues() 

Anyways - I don't really mind leaving as is, it doesn't affect anything much, would love your opinion :)

michaelavila:
@freak4pc :+1: Let me know what you think. I made the explanations more consistent, changed up some of the wording, and renamed the tests (along with the reording you asked for).

michaelavila:
@freak4pc my pleasure. Thanks for the quick responses and feedback. Hopefully I have more to contribute. Take care.

PR #137, opened 2018-02-17
toSortedArray operator

jdisho:
This pull request proposes a way to transform an observable of comparables into an observable of ordered arrays.

### DIFF
--- a/Source/RxSwift/sorted.swift
+++ b/Source/RxSwift/sorted.swift
@@ -1,0 +1,35 @@
+//
+//  Sorted.swift
+//  RxSwiftExt
+//
+//  Created by Joan Disho on 17.02.18.
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import Foundation
+import RxSwift
+
+extension Observable where E: Sequence, E.Iterator.Element: Comparable {
+    
+    /**
+     Transforms an observable of comparables into an observable of ordered arrays.
+     
+     - returns: The sorted observable.
+    */
+    
+    public func sorted<T>() -> Observable<T> where E.Iterator.Element == T {
+        return self.map { $0.sorted() }
+    }
+    
+    /**
+     Transforms an observable of comparables into an observable of ordered arrays by using the function passed in.
+     
+     - parameter areInIncreasingOrder: A function to compare to elements.
+     - returns: The sorted observable.
+     */
+    
+    public func sorted<T>(_ areInIncreasingOrder: @escaping (T, T) -> Bool) -> Observable<[T]> 
+        where E.Iterator.Element == T {
+        return self.map { $0.sorted(by: areInIncreasingOrder) }
+    }
+}
--- a/Source/RxSwift/sorted.swift
+++ b/Source/RxSwift/sorted.swift
@@ -1,5 +1,5 @@
 //
-//  Sorted.swift
+//  sorted.swift
 //  RxSwiftExt
 //
 //  Created by Joan Disho on 17.02.18.
--- a/Source/RxSwift/sorted.swift
+++ b/Source/RxSwift/sorted.swift
@@ -13,12 +13,20 @@ extension Observable where E: Sequence, E.Iterator.Element: Comparable {
     
     /**
      Transforms an observable of comparables into an observable of ordered arrays.
-     
+
+     - default: ascending order
      - returns: The sorted observable.
     */
     
-    public func sorted<T>() -> Observable<T> where E.Iterator.Element == T {
-        return self.map { $0.sorted() }
+    public func toSortedArray<T>(
+        ascending: Bool = true
+        ) -> Observable<[T]> where E.Iterator.Element == T {
+        guard ascending else {
+            return self.map { $0.sorted(by: >) }
+        }
+
+        return self.map { $0.sorted(by: <) }
+
     }
     
     /**
@@ -28,8 +36,9 @@ extension Observable where E: Sequence, E.Iterator.Element: Comparable {
      - returns: The sorted observable.
      */
     
-    public func sorted<T>(_ areInIncreasingOrder: @escaping (T, T) -> Bool) -> Observable<[T]> 
-        where E.Iterator.Element == T {
-        return self.map { $0.sorted(by: areInIncreasingOrder) }
+    public func sorted<T>(
+        _ using: @escaping (T,T) -> Bool
+        ) -> Observable<[T]> where E.Iterator.Element == T {
+        return self.map { $0.sorted(by: using) }
     }
 }
--- a/Source/RxSwift/toSortedArray.swift
+++ b/Source/RxSwift/toSortedArray.swift
@@ -2,37 +2,33 @@
 //  toSortedArray.swift
 //  RxSwiftExt
 //
-//  Created by Joan Disho on 17.02.18.
-//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//  Created by Joan Disho on 17/02/18.
+//  Copyright © 2018 RxSwift Community. All rights reserved.
 //
 
 import Foundation
 import RxSwift
 
-extension Observable where E: Sequence, E.Element: Comparable {
+public extension ObservableType {
     /**
-     Transforms an observable of comparables into an observable of ordered arrays.
+     Converts an Observable into another Observable that emits the whole sequence as a single array, sorts it using the provided closure and then terminates.
 
-     - parameter ascending: Defined the sorting direction. Defaults to ascending.
-     - returns: The sorted observable.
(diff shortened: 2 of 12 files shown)


freak4pc:
It looks interesting, but given it would only be relevant for a finite stream (e.g. only after it completex), can you explain the use case for it?

freak4pc:
Sounds like a good middle ground @fpillet - 
@jdisho When you have the time, make the changes and we'll review again. No rush! :) 

Thanks !

Source/RxSwift/sorted.swift
```diff
@@ -28,8 +36,9 @@ extension Observable where E: Sequence, E.Iterator.Element: Comparable {
      - returns: The sorted observable.
      */
     
-    public func sorted<T>(_ areInIncreasingOrder: @escaping (T, T) -> Bool) -> Observable<[T]> 
-        where E.Iterator.Element == T {
-        return self.map { $0.sorted(by: areInIncreasingOrder) }
+    public func sorted<T>(
+        _ using: @escaping (T,T) -> Bool
```
freak4pc: The parameter should be `by areInIncreasingOrder: @escaping (T, T) -> Bool`, much like your original documentation and also like the original sorted method.

Source/RxSwift/sorted.swift
```diff
@@ -13,12 +13,20 @@ extension Observable where E: Sequence, E.Iterator.Element: Comparable {
...
+     - default: ascending order
      - returns: The sorted observable.
     */
     
-    public func sorted<T>() -> Observable<T> where E.Iterator.Element == T {
-        return self.map { $0.sorted() }
+    public func toSortedArray<T>(
+        ascending: Bool = true
+        ) -> Observable<[T]> where E.Iterator.Element == T {
+        guard ascending else {
```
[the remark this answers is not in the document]
jdisho: I like this! :D

jdisho:
Hey @freak4pc,
Sorry for the delay! You can review it now 😊

Source/RxSwift/toSortedArray.swift
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     Transforms an observable of comparables into an observable of ordered arrays by using the function passed in.
+     
+     - parameter areInIncreasingOrder: A function to compare to elements.
+     - returns: The sorted observable.
+     */
+    public func toSortedArray<T>(
+        _ areInIncreasingOrder: @escaping (T, T) -> Bool
+        ) -> Observable<[T]> where E.Iterator.Element == T {
+        return self.map { $0.sorted(by: areInIncreasingOrder) }
```
freak4pc: drop the `self.`

Source/RxSwift/toSortedArray.swift
```diff
@@ -0,0 +1,37 @@
...
+extension Observable where E: Sequence, E.Iterator.Element: Comparable {
+    /**
+     Transforms an observable of comparables into an observable of ordered arrays.
+
+     - parameter ascending: Defined the sorting direction. Defaults to ascending.
+     - returns: The sorted observable.
+    */
+    public func toSortedArray<T>(
+        ascending: Bool = true
+        ) -> Observable<[T]> where E.Iterator.Element == T {
```
freak4pc: Can we change the signature to have ascending in-line?  And also make the same Element fix.
e.g.

```swift
public func toSortedArray<T>(ascending: Bool = true)
    -> Observable<[T]> where E.Element == T {
```

Source/RxSwift/toSortedArray.swift
```diff
@@ -0,0 +1,37 @@
...
+//  RxSwiftExt
+//
+//  Created by Joan Disho on 17.02.18.
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import Foundation
+import RxSwift
+
+extension Observable where E: Sequence, E.Iterator.Element: Comparable {
```
freak4pc: `E.Iterator.Element` -> `E.Element`

freak4pc:
Also noticed something @jdisho 
If we want to match `toArray` - > that one emits only when the sequence completes.

Your `toSortedArray` will re-sort the array whenever there's a new emission - meaning if the array changes, you'd get a different sort. I'm not necessarily sure if this is correct or needed, even though I could see it being somewhat useful in some cases.

@fpillet would really appreciate your thought here if you have a second.

freak4pc:
@jdisho Yeah - I do understand what you're saying, the problem here is as I said, every time a new item is emitted, you will sort the array again completely. 

`toArray`, for example, only converts to an array when the sequence completely finished emitting. Also, your version works on an `Observable<Sequence>`, where `toArray` works on any observable (and "makes it" into an array as it completes).

So I think if we use the same naming (`.toSortedArray`) , the behavior doesn't make sense. 

The way you implemented your operator means the following will happens: 

You emit [2], you get [2]
You emit [1, 2], you get [1, 2]
you emit [-5, 1, 2] you get [-5, 1, 2]

Every emissions causes a new sorting, which I'm not sure is the planned behavior here (also your tests use a finite sequence, e.g. a complete one). 

If we want to get parity with `toArray` behavior-wise, the implementation would be more similar to this:
```swift
public extension ObservableType where E: Comparable {
    func toSortedArray(ascending: Bool = true) -> Observable<[E]> {
        return toSortedArray(by: { ascending ? $0 < $1 : $0 > $1 })
    }
}

public extension ObservableType {
    func toSortedArray(by: @escaping (E, E) -> Bool) -> Observable<[E]> {
        return toArray().map { $0.sorted(by: by) }
    }
}
```

Any thoughts from anyone on @RxSwiftCommunity/contributors ? 
Specifically @fpillet as you mentioned the `toSortedArray` naming. If we want sorting on every emission the naming doesn't really fit, and if we want the naming then it should only happen on completion IMO.

jegnux:
I agree for the naming. `toSortedArray()` should emit the sorted array only on completion. 
Maybe we can have a version that emit a sorted array on each source emission named `scanAsSortedArray()`? I think that `scan` helps to understand that it's an accumulator.

freak4pc:
@jegnux It's not really an accumulator though - its just a simple map, think about the fact it doesn't care about the "past" (unlike `scan`). 

I think `toSortedArray()` should be the completed version and if we want a per-emission sorting we should just call it `sorted` or `sort` (like `map`) IMO.

freak4pc:
Yup I agree with you that the its very confusing at the moment. 
The second one should be `toSortedArray`. 

If the first one is a different variation its a bit tricky, I don't like using scan in operators because it means its a side-effect you can't "get out" of (you can't really end the accumulation easily). So we need to be careful in using it. 

The implementation @jdisho has right now does this: 

----[1]------[8, 1]--------[8, 1, 2]------|--->
------[1]-------[1, 8]---------[1,2,8]---|--->

jdisho:
@freak4pc Ahh.. now I got the problem.  Than yeah, if it is suppose to emit only in completion that I should make the suggested adaptations :) 

```swift
public func toSortedArray<T>(_ areInIncreasingOrder: @escaping (T, T) -> Bool)
        -> Observable<[T]> where E.Element == T {
            return toArray().map { $0.flatMap { $0.sorted(by: areInIncreasingOrder) } }
    }
```

freak4pc:
Hey @jdisho - I made a few changes to the implementation. 

I think the original intent was lost so I wanted to clean it up juts a tad. You implementation was still working only for Array types, where `toArray` works for any Observable.

The usage of `toArray` and `toSortedArray` should be similar - e.g. a finite sequence of individual elements is converted to an array and sorted. 

Let me know if this makes sense to you !

jdisho:
Thank you @freak4pc  😄

freak4pc:
Gonna leave this open for a day or two for any feedback from @RxSwiftCommunity/contributors and will merge then :)

Source/RxSwift/toSortedArray.swift
```diff
@@ -2,37 +2,33 @@
...
-    public func toSortedArray<T>(_ areInIncreasingOrder: @escaping (T, T) -> Bool)
-        -> Observable<[T]> where E.Element == T {
-            return toArray().map { elements in
-                elements.flatMap { element in
-                    element.sorted(by: areInIncreasingOrder)
-                }
-            }
+     Converts an Observable into another Observable that emits the whole sequence as a single array, sorts it using the provided closure and then terminates.
+
+     - parameter ascending: Should the emitted items be ascendign or descending.
```
freak4pc: ascendign -> ascending 

_reminder to myself_

jegnux:
I'm fine with the naming and implementations 👍 

What about the version which emits for every source element emitted?
I think it would be great to have such equivalents for both `toArray()` and `toSortedArray()`.

But maybe it's kinda out of scope for this PR.

freak4pc:
Thanks @jdisho !

PR #138, opened 2018-02-20
Implement toggle operator

vincent-pradeilles:
This operator can be particularly useful when working with buttons through RxCocoa.


freak4pc:
Hey @vincent-pradeilles , Many thanks for your contribution. 

I kinda feel this operator is a bit too tiny to be added but maybe I would be the only one thinking this. 

@fpillet Could you share your opinion? It seems like such a tiny surface area (just wrap scan, practically) that I'm not sure it's worth it's own operator

vincent-pradeilles:
Hello @freak4pc 

I actually made the PR because I had to use this operator today, and I felt that, without it, having the combination of `scan()` and `startWith()` in a controller didn't felt really expressive.

Of course I'm not sure if the goal of this project is strictly towards operator that bring significant, or if it's also open to ones that just help improve readability.

Anyway, I'll let the community decide ;)

fpillet:
This is an interesting operator -- a bit focused on a narrow topic, but that's the purpose of RxSwiftExt after all. The only thing I'd change is the name -- should find a name more explicit for what the operator does. Maybe `toggleOnNext` or `toggleEach`? @freak4pc any better suggestion ?

Thinking about this, it may be a good opportunity to build a more general `onEach` operator. `.toggleBool` could be one of the cases, but others would be interesting as well like `.pool(count: Int)`, etc. I feel there is an opportunity to generalize the functionality here ... 

Thoughts ?

freak4pc:
@vincent-pradeilles In general I agree, but in this case, since it wraps scan which is a bit of a "side-effecty" operator, some things are happening while the user wouldn't understand necessarily what's the underlying implementation of it. Or the fact that scan creates sort of a "state" store. 

@fpillet I like the possibility of a onEach - but onEach is practically `map`, right? 

If we add this toggling operator, I would at least try to contain it to `Bool` ObservableTypes to begin, with, e.g. 

```swift
extension ObservableType where E == Bool {
  func toggle() -> Observable<Bool> {
    return flatMapLatest {
      return self.scan($0) { _, val in return !val }
    }
  }
}
```

freak4pc:
@fpillet @vincent-pradeilles 

Thinking about this a few more minutes, what would make sense here is either name it 

`toggleBool(initial:)` or `toggleBool(seed:)` (seed like reduce or scan)

IMO

vincent-pradeilles:
@fpillet on my side, I find the idea of a `.pool(count: Int)` operator kind of nice. It could even be generalized to `.pool(iterator: Iterator)`, in order to have a kind of round-robin set of values to circulate.

Anyway, I'm completely okay to implement them I there is a consensus ;)

fpillet:
Thinking about this more, I'm not so sure about `.pool(count: Int)` anymore since RxSwift already has the `window` operator. Also, the functionality is very much linked to the particular application that will use it.

Maybe the basic `toggle` is simple enough after all :)

freak4pc:
If we call it toggle, lets be specific with `toggleBool(initial:)` or `toggleBool(seed:)`. A general Bool is confusing unless you limit it to `ObservableType where E == Bool`, but that's very limiting and probably useless :) 

Makes sense @vincent-pradeilles ?

freak4pc:
Any extra toggle ? Are we ok with toggleBool(initial:) or some other variation ?

freak4pc:
@vincent-pradeilles @fpillet Would love your feedback if you have some time so we can push this forward :)

freak4pc:
Should we close this for now or are we still interested in this operator?

vincent-pradeilles:
@freak4pc `toggleBool(initial:)` might be an interesting API for the operator. Personally I still believe that this operator holds some value, especially when working with UI.

fpillet:
Lets get it in! As Vincent says, it can be useful when working with UI.

vincent-pradeilles:
If everyone feels comfortable with the API `toggleBool(initial:)`, I'll make the changes at the end of the day ;)

freak4pc:
Yup I think we're good with that. Make the changes and I'll review them :) 
Thanks ! 

PS don't forget to rebase

vincent-pradeilles:
Changes have been made

Readme.md
```diff
@@ -546,6 +547,33 @@ next((Test Class, 13))
...
+    
+Observable.from(input)
+    .toggleBool(initial: false)
+    .subscribe(onNext: { result in
+        print(result)
+    })
+```
+
+```
+Input: [1, 2, 3, 4]
```
freak4pc: Can you align the output with other operators, e.g. just print 

next(false)
next(true)
next(false)
next(true)
next(false)
completed

Source/RxSwift/toggleBool.swift
```diff
@@ -0,0 +1,24 @@
...
+    
+    /**
+     Returns an observable sequence of booleans, starting with `initial` and toggling its value
+     at each `.next` event
+     
+     - parameter initial: the boolean value to initialize the sequence
+     - returns: An observable sequence of booleans that starts with `initial` and toogles its value at each `.next` event
+     */
+    public func toggleBool(initial: Bool) -> Observable<Bool> {
+        return scan(initial) { previous, _ in return !previous }.startWith(initial)
```
freak4pc: Implementation question for both yourself and @fpillet 
I think the startWith shouldn't be part of this operator necessarily, since scan itself has a seed value. 

It creates a scenario that by providing 4 input values, ill get 5 output emissions 

e.g. 1 2 3 4 (seed = false)
emit: false true false true false
Where the first two would happen together.

I'd expect it to be just `false, true, false, true` while the first one emits only upon the emitting of the source sequence. What do you guys think? Does either make more sense than the other?

Tests/RxSwift/ToggleBoolTests.swift
```diff
@@ -0,0 +1,56 @@
...
+import RxSwift
+import RxTest
+import RxSwiftExt
+
+class ToggleBoolTests: XCTestCase {
+    
+    func testStartWithFalse() {
+        let originalSequence = Observable.of(1, 2, 3, 4)
+        let correct: [Recorded<Event<Bool>>] = [
+            next(0, false),
```
freak4pc: I think I would prefer testing timed events instead of all events being false, so we can confirm the time of the emissions are respected. You can use createColdObservable in RxTest to do that.

vincent-pradeilles: Done

Readme.md
vincent-pradeilles: Done

Source/RxSwift/toggleBool.swift
vincent-pradeilles: To me it made sense to have the initial value, because I was working with UI, so I expected the resulting Observable to have an initial value, like a Behavior.

freak4pc: But would you expect an operator to emit something without any values going through it? It seems confusing to me. It means an empty sequence will still emit the initial value.

vincent-pradeilles: If the function were to return an `BehaviorSubject` instead of an `Observable`, would this behavior then be deemed more acceptable?

MortyMerr:
Hey @freak4pc could u please review it? If the implementation is ok, I am ready to resolve conflicts, cause this operator seems very suitable for my needs

Source/RxSwift/toggleBool.swift
freak4pc: I think we can change this to `reduce(into: initial) { previous, _ in previous.toggle() }`

Thoughts?

MortyMerr: `reduce` emits only final value. Are you sure such behavior is required?

freak4pc: meant scan into.

MortyMerr: It doesn't emit initial value

MortyMerr: Actually, I just wonder why `startWith` needed at all. I don't want sequence to emit something before source

freak4pc: I think you're missing my point. I wanted to use scan(into) and `toggle`. The `startWith` should still be there.

MortyMerr: I mean common example will be like:
```
button.rx.tap.toggleBool(initial: true).bind(to: view.rx.isHidden)
```
And i want view becomes hidden or not only after first button tap

MortyMerr: > use scan(into) and toggle
What's the difference with the current? Quite subjective IMO. But yes, can be implemented in that way.

MortyMerr: Is that all? I can create new PR with that change

MortyMerr: > But would you expect an operator to emit something without any values going through it? It seems confusing to me. It means an empty sequence will still emit the initial value.

Fully agree

freak4pc: Yup, it's been a long time since we discussed this one. I don't think it needs an initial value.

freak4pc: I just think this is is mirroring Foundation's toggle, so it's nicer to use that. Yes, it's quite subjective.

freak4pc: TBH I can't really say this operator had too much community interest. Evidently only a few people (2-3) have asked for it and it's really minor. Don't mind adding it since it's in Foundation but I think it's quite minor anyways.

MortyMerr: If it's not popular/needed - it's better to close it.

MortyMerr:
Hey @freak4pc what do u think about that? Close or merge (after conflicts resolving)? :) I am ok with both answers

freak4pc:
As of now due to the lack of community interest and no ownership here I think it would be better to close it. Thoughts?

vincent-pradeilles:
@freak4pc agreed on my part 🙂

PR #139, opened 2018-02-21
Adds filterMap(as:) operator

jegnux:
## `filterMap(as:)`
This PR adds a `filterMap(as:)` operator to ObservableType.
It allows to cast sequence elements as the given type and filter out any element that can be casted.

Other possible names for this operator could be:
- `compactMap(as:)` to match new Swift 4 `compactMap` method on Collection
- `map(as:)` to stay simple and match `map(to:)` and `map(at:)`, but it's not clear that elements can be filtered

### DIFF
--- a/Source/RxSwift/filterMapAs.swift
+++ b/Source/RxSwift/filterMapAs.swift
@@ -1,0 +1,26 @@
+//
+//  filterMapAs.swift
+//  RxSwiftExt
+//
+//  Created by Jérôme Alves on 21/02/2018.
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import RxSwift
+
+extension ObservableType {
+
+    /**
+     Takes a sequence of elements and returns a sequence of elements casted as given type, filtering out any non castable values.
+
+     - parameter type: A type whose matches the element type of the input sequence
+     - returns: An observable sequence containing the elements 
+     */
+    public func filterMap<R>(as type: R.Type) -> Observable<R> {
+        return self.filterMap {
+            guard let result = $0 as? R else { return .ignore }
+            return .map(result)
+        }
+    }
+
+}
--- a/Source/RxSwift/filterMapAs.swift
+++ b/Source/RxSwift/filterMapAs.swift
@@ -11,16 +11,23 @@ import RxSwift
 extension ObservableType {
 
     /**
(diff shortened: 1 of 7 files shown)


freak4pc:
There is already a PR to do this here - #134 

I suggested there and I think I still think the same, that naming similar to `filterType` or `takeType` would be better (IMO).

It could return it if the type matches, or try to cast it and return it, and otherwise just not emit. 

compactMap is more generic to nils generally, IMO ?

Source/RxSwift/filterMapAs.swift
```diff
@@ -0,0 +1,26 @@
...
+     Takes a sequence of elements and returns a sequence of elements casted as given type, filtering out any non castable values.
+
+     - parameter type: A type whose matches the element type of the input sequence
+     - returns: An observable sequence containing the elements 
+     */
+    public func filterMap<R>(as type: R.Type) -> Observable<R> {
+        return self.filterMap {
+            guard let result = $0 as? R else { return .ignore }
+            return .map(result)
+        }
```
jegnux: There are many different implementation possible for this operator. I chose this one to explicitly match the `filterMap` in the name.

Other possible implementations: 
```swift
return self.filter { $0 is R }.map { $0 as! R }
```
```swift
return self.map { $0 as? R }.unwrap()
```
```swift
return self.flatMap { ($0 as? R).map(Observable.just) ?? .empty() }
```
And I'm pretty sure there are more...

I don't know if an implementation is more optimized than another. 
If it does, feel free to tell me it and I'll update the code.

jegnux:
@freak4pc oh ! My bad, I should have looked to open PR.
What should I do? Close this one and discuss on other one?
What about deprecations? Should I open a new PR with only those changes?

freak4pc:
@jegnux For deprecations yeah, totally lets make a new PR.

For this one, I think its better to unify our efforts, but for now don't close this - the author of the other one is not responding, so if this won't move in a new direction in the next few days, let's pursue this separately. What do you think?

jegnux:
I'm ok. For the "naming" debate, I answered on the other PR as the debate started there.

jegnux:
✅ I removed deprecations from this PR and opened a separated PR instead https://github.com/RxSwiftCommunity/RxSwiftExt/pull/140

freak4pc:
Closed in favor of #134

PR #140, opened 2018-02-21
Deprecate mapTo() and mapAt() in favor to map(to:) and map(at:)

jegnux:
I noticed that `mapTo` was deprecated in favor to `map(to:)` on `SharedSequenceConvertibleType` (in **mapTo+RxCocoa.swift**), but this deprecation wasn't applied to `mapTo()` implementation on `ObservableType`. 
I thought it was great to uniforms this. I also deprecated new `mapAt()` in favor of `map(at:)` to keep some consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+- deprecated `mapTo()` and `mapAt()` in favor of `map(to:)` and `map(at:)`
 - added `mapAt(keyPath:)` operator
 - added `zip(with:)` operator
 
--- a/Playground/RxSwiftExtPlayground.playground/Pages/mapAt.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/mapAt.xcplaygroundpage/Contents.swift
@@ -13,9 +13,9 @@ import RxSwift
 import RxSwiftExt
 
 /*:
- ## mapAt(KeyPath)
+ ## map(at: KeyPath)
 
- The `mapAt` operator transforms a sequence of elements where each element is mapped to its value at the provided key path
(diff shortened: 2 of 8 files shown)


freak4pc:
@jegnux I'm actually confused - I think we moved from `map(at:)` back to `mapAt()` to avoid overloads as these were causing conflicts with the main libraries (also, overloads are costly in performance). 

@fpillet Can you confirm if you have a minute? I remember we moved away from overloads, so I think this PR is going the other way.

jegnux:
@freak4pc oh, now I understand. You moved back from `map(to:)` to `mapTo()` but you forgot to do it in `SharedSequenceConvertibleType`. That's why I was confused.

If you already debate about this, let's close this PR, but I have to admit I'm a little bit sad, `map(to:)` and `map(at:)` fit way better in Swift, and with recent RxSwift changes (`bindTo()` => `bind(to:)`, `disposedBy()` => `disposed(by:)`, etc...), and with our own operators (`zip(with:)`).

Should we really go against (Rx)Swift API design guidelines to workaround performance issues? And what are those performance issues? at compilation or at runtime?

freak4pc:
The issue wasn't performance - there were overload collisions with RxSwift itself (e.g. Swift gets confused with the regular map(s) )

Regrading performance - type inference is costly, so overloads are costly :) 
I would also prefer map(to:) but florent changed it a while back due to conflicting issues IIRC.

fpillet:
Sorry guys, been handling emergencies all day. Not a minute to spare looking at this. My opinion on `map` has not changed: *never* name an operator `map`, no matter how smart your parameter is named. This is simply a recipe for disaster as it will inevitably conflict, one way of the other, with the `map {` trailing closure form.

PR #144, opened 2018-02-26
Soften RxSwift dependency for Carthage

clementleroy:
Just a small adjustment to the RxSwift required version. It should now use the latest version of RxSwift.


freak4pc:
Thanks Clemenet! Great fix 👍

freak4pc:
^ @ashfurrow any ideas? I guess its just a one off

ashfurrow:
Hmm, that doesn't look good. Not sure what's going on, and there are no useful server logs. Let's keep an eye on it.

PR #146, opened 2018-03-23
Improve on ObservableType+Weak to offer a nicer call syntax

vincent-pradeilles:
There are two things in this PR:

- a couple of wrapper functions around the `subscribe` functions already defined in `ObservableType+Weak.swift` that allow for cleaner syntax at call sites (cf. playground page and README.md)
- a `flatMap` function that follows the same principle of encapsulating weak references. This one can be particularly useful with MVVM architecture, which often have needs of `[weak self]` when flatMapping within Controllers or ViewModels


freak4pc:
Hey there - can you please provide a description in your PR to explain what you’re doing here ?

vincent-pradeilles:
There are two things in this PR:

- a couple of wrapper functions around the `subscribe` functions already defined in `ObservableType+Weak.swift` that allow for cleaner syntax at call sites (cf. playground page and README.md)
- a flatMap function that follow the same principle of dealing with weak references. This one can be particularly useful with MVVM architecture, which often have needs of `[weak self]` when flatMapping within Controllers or ViewModels

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -47,6 +49,7 @@ extension ObservableType {
                 return Observable<O.E>.error(error)  as! O
```
freak4pc: Couldn't you do 

```swift
let error: O = Observable<O.E>.error(error)
return error
```

Just to satisfy the constraint?

freak4pc:
@vincent-pradeilles Thanks for this ! Can you provide some sample usage in the PR body itself to help reviewers understand the benefits of these helpers ?

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: No, you'd still have to cast in order to have the affectation compile.

vincent-pradeilles:
@freak4pc sure, he are some additional use cases:

```swift
enum SampleError: Error {
    case fatalError
}

class Foo {
    let obs1 = Observable.of(3, 4, 6, 7)
    let obs2 = { (value: Int) in Observable.of(value * 2, value * 3, value * 4) }
    let obs3 = Observable<String>.create { observer in
        observer.onNext("First")
        observer.onError(SampleError.fatalError)
        observer.onNext("Second")
        observer.onCompleted()
        return Disposables.create()
    }

    let disposeBag = DisposeBag()

    init() { }

    deinit { print("deinit") }

    func display(_ value: Any) {
        print(value)
    }

    func work1() {
        obs1.subscribe(weak: self) { strongSelf, event in
            strongSelf.display(event)
        }.disposed(by: disposeBag)
    }

    func work2() {
        obs1.flatMap(weak: self) { (strongSelf, value) in
            return strongSelf.obs2(value)
        }.subscribe(weak: self) { strongSelf, event in
                strongSelf.display(event)
        }.disposed(by: disposeBag)
    }

    func work3() {
        obs3.subscribe(weak: self) { (strongSelf, event) in
            strongSelf.display(event)
        }
    }
}
```

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -83,6 +119,33 @@ extension ObservableType {
...
+     - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence..
+     - parameter onDisposed: FAction to invoke upon any type of termination of sequence (if the sequence has gracefully completed, errored, or if the generation is canceled by disposing subscription).
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     */
+    public func subscribe<A: AnyObject>(weak obj: A,
+                                        onNext: ((A, E) -> Void)? = nil,
+                                        onError: ((A, Error) -> Void)? = nil,
+                                        onCompleted: ((A) -> Void)? = nil,
+                                        onDisposed: ((A) -> Void)? = nil) -> Disposable {
+        let curriedOnNext: ((A) -> (Self.E) -> Void) = { (strongObj: A) in { (value: E) in onNext?(strongObj, value) } }
```
freak4pc: very cool! :) 
Currying always sort of looks like a nightmare to me but here this is used very smartly ;) 

It's a bit confusing though. Wonder if there's any good way to break it up a bit, by at least moving the curried blocks to their own lines. What do you think ?

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -83,6 +119,33 @@ extension ObservableType {
...
 	}
+    
+    /**
+     Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.
+     
+     - parameter weak: The object that we want to have a safe and unretained reference on.
+     - parameter onNext: Action to invoke for each element in the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence..
+     - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence..
+     - parameter onDisposed: FAction to invoke upon any type of termination of sequence (if the sequence has gracefully completed, errored, or if the generation is canceled by disposing subscription).
```
freak4pc: `FAction -> Action` , but also in general, can we use the same docs from the original `subscribe(weak)` ? It uses wording like function, vs. action

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -83,6 +119,33 @@ extension ObservableType {
 
 		return Disposables.create(self.asObservable().subscribe(observer), disposable)
 	}
+    
+    /**
+     Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.
```
freak4pc: Let's add at the end here:

> while providing a safe, unretained reference to.

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -83,6 +119,33 @@ extension ObservableType {
...
+     
+     - parameter weak: The object that we want to have a safe and unretained reference on.
```
freak4pc: on -> to

Playground/RxSwiftExtPlayground.playground/Pages/weak.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,49 @@
...
+    func display(_ value: Any) {
+        print(value)
+    }
+    
+    func produce(_ value: Int) -> Observable<Int> {
+        return Observable.of(value * 2, value * 3)
+    }
+}
+
+example("weak") {
```
freak4pc: I wonder if we should make a page for each of these operators separately instead of sticking them in a single page. We usually have a page-per-operator, it seems.

Readme.md
```diff
@@ -511,6 +512,49 @@ completed
...
+    func produce(_ value: Int) -> Observable<Int> {
+        return Observable.of(value * 2, value * 3)
+    }
+}
+
+let observable = Observable.of(2, 5, 7)
+
+var tool: Tool? = Tool()
+    
+observable.flatMap(weak: tool!) { (strongTool, value) in
```
freak4pc: We shouldn't use force unwrap in examples... the examples should provide relatively clean standards of usage so they don't confuse a developer.

In general might be nice to put some of the examples you put in the PR itself instead. They seemed to be much more descriptive than these and more "real-life".

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -29,6 +29,28 @@ extension ObservableType {
...
+    public func flatMap<A: AnyObject, O: ObservableConvertibleType>(weak obj: A, _ selector: @escaping (A, Self.E) throws -> O) -> Observable<O.E> {
+        // swiftlint:disable force_cast
+        // Force casts are unavoidable here, because of Swift's type system limitation
+        return flatMap { [weak obj] value -> O in
+            guard let strongObj = obj else { return Observable<O.E>.empty() as! O }
+            
+            do {
+                return try selector(strongObj, value)
+            } catch {
+                return Observable<O.E>.error(error)  as! O
```
freak4pc: two spaces before `as!` should be a single space.

Source/RxSwift/ObservableType+Weak.swift
```diff
@@ -29,6 +29,28 @@ extension ObservableType {
             method?(obj)()
         }
     }
+    
+    /**
+     Performs a flatMap providing a safe unretained reference to `obj` as an argument of the flatMap closure
```
freak4pc: Let's use the description of original flatMap, something like 

> Projects each element of an observable sequence to an observable sequence and merges the resulting observable sequences into one observable sequence, while providing a safe & unretained reference to the provided object.

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: I'm not sure I understand what you mean by "moving the curried blocks to their own lines".

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: Done

Source/RxSwift/ObservableType+Weak.swift
freak4pc: Nevermind, re-read this. It's fine :)

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: done

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: done

Readme.md
vincent-pradeilles: done

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: done

Source/RxSwift/ObservableType+Weak.swift
vincent-pradeilles: done

Tests/RxSwift/WeakTests.swift
```diff
@@ -133,4 +133,40 @@ class WeakTests: XCTestCase {
...
+        let value: T
+        
+        init(value: T) {
+            self.value = value
+        }
+    }
+    
+    private var wrapped: Wrapper<Int>? = nil
+    
+    func testFlatMap() {
```
freak4pc: I think we're missing tests for your new subscribe overload ?

Playground/RxSwiftExtPlayground.playground/Pages/weak.xcplaygroundpage/Contents.swift
vincent-pradeilles: done

Tests/RxSwift/WeakTests.swift
vincent-pradeilles: Regarding that one, I was not sure of the testing strategies to have do the overloads.

freak4pc:
Just thinking out loud and would love your opinion. I think it might be better to name the subscribe one -  `subscribeSafe` or `subscribeUnretained` rather then `subscribeWeak`. Two reasons: 

1. Overloads are tricky for performance and also stability sometimes
2. What you get in the `onNext` block isn't weak/optional, it's already "Strongified" after is has been weakified and you got a safe reference to it. 

Does that make sense? `subscribeSafe` sounds kinda cool TBH.

vincent-pradeilles:
`subscribeSafe` makes sense to me, I'll wait a little and look at what people propose before I make the change ;)

groue:
Hello, here are my two cents about "safe".

First cent: I recommend everyone to subscribe to the [Swift Forums](http://forums.swift.org). One quickly learns there that "safe", in Swift lingo, does not mean "don't crash". Instead, "safe" often means "crash if preconditions are not met". For example, according to this terminology, integer addition is "safe" because it traps on overflow. I'm not here to tell if this terminology is good or not, expected or not. I'm just here to try to avoid redefining an overloaded term.

Second cent: when I read `observable.subscribe(safe: self, onNext: { ... })`, I don't quite understand what is made "safe". And I even start to wonder if all other subscription methods are unsafe (without knowing how). I don't think this is your goal :-)

freak4pc:
@groue That's a fair point. Do you have a thought on how to name a method that provides this shortcut? `subscribe(unretaining:)` (or unretained:) is more descriptive but I'm worried its a bit confusing as well. 

Any thoughts would be very welcomed :)

groue:
`unretainedContext` or `unretaining` are really to the point, and I find the following code snippet crystal-clear:

```swift
obs.flatMap(unretainedContext: self) { (strongSelf, value) in
    return strongSelf.obs2(value)
}.subscribe(unretainedContext: self) { (strongSelf, event) in
    strongSelf.display(event)
}.disposed(by: disposeBag)
```

The general pattern (without `strongSelf`) is less clear - but still recognizable (and super handy!!):

```swift
obs.flatMap(unretaining: val) { (val, value) in
    return val.obs2(value)
}.subscribe(unretaining: val) { (val, event) in
    val.display(event)
}.disposed(by: disposeBag)
```

More eyes may help. But "safe"... I just don't know what kind of hazard it protects me from.

freak4pc:
I'm ok with `unretaining:` . `unretainedContext` is too much of a mouthful and also really spooky-looking :D 

@vincent-pradeilles Thoughts?

bobgodwinx:
How about `subscribe(unretained: obj...)` ?

freak4pc:
@bobgodwinx I'm sort of OK with it. The only worry is that it might make people thing they need to provide an unretained object, which isn't really the right message.

bobgodwinx:
This is clear that the object is not retained. + there is also documentations and comments on the method signature. I am for the shorter the better. :) swift-like

freak4pc:
It's just a single-letter difference :)  (unretaining vs. unretained)

I kinda prefer unretained as its reads like a sentence, which I'm always a fan of.

`subscribe(unretained: self ...` reads subscribe to unretained self, which is kind of cool, IMO.

freak4pc:
@bobgodwinx I edited my comment, I meant the other way around haha :)

jegnux:
@vincent-pradeilles I'm not fully convinced by these operators... They bypass a language "feature/constraint" in a closed way. Here you have implemented this feature on `subscribe` and `flatMap`, but to be consistent, it should be available on every operator that takes closure. And there's a lot: `create`, `deferred`, `do`, `takeWhile`, `takeUntil`, `map`, `combineLatest`, `withLatestFrom`, etc... 

🤔  Moreover, please @RxSwiftCommunity/contributors correct me if I'm wrong, but I don't feel this kind of convenience operator have their place in **RxSwiftExt**? Because they're not really new operators that have a **strong and distinct meaning and behavior**. They "just" handle the responsibility to weakify/strongify a reference. 

💡  What about making a new **RxSwiftCommunity** repository named **RxSafe**, **RxWeak**, or something like that (matching naming convention chosen above)? IMHO it would be a better place and will encourage implementation of other closure-based operators mentioned in my first point.

jegnux:
Oh, I just noticed that **RxSwiftExt** already includes a `subscribe(weak:_:)` and @vincent-pradeilles is actually bringing it to `flatMap`. As it wasn't documented in README, I never payed attention to it. 

This makes me believe even more that it should be extracted in its own repository in order to encourage implementation of this pattern on more operators and to make easier they discoverability.

freak4pc:
@jegnux I'd love other opinions but I don't agree personally. 
We can go for an approach of subspec-ing that piece but I still think its an overkill - 

RxSwiftExt doesn't have a "mission statement" of introducing _only_ new operators. Actually the repo is just about extending the existing abilities of RxSwift itself, with things that don't necessarily fit within the main repo. 

For that reason, I specifically think it fits just fine in RxSwiftExt.

Yes, we do need to introduce this ability in other operators as well, but it would probably be only where its extremely useful, and not just to have full coverage. For example it's very useful in a `map` (since it saves opening a full closure vs. using a one-liner), but its not super useful with `create`, `deferred` or `withLatestFrom` (which I don't entirely understand why you'd even need to un-retain an object in these scenarios)

We now have two thoughts on both sides, and I'd love to hear more from the community - people are a bit busy this time of year so I hope we'll get some more opinions going :)

Tests/RxSwift/WeakTests.swift
freak4pc: How do you mean exactly ?

fpillet:
@jegnux discoverability is mainly because the original contributor didn't add it to the README and since I found it a bit difficult to use (but maybe because of my fault) I left it "hidden".

This said, as @freak4pc puts it we are not bound to propose only operators. RxSwiftExt's goal is to be a grab-bag of things Kruno doesn't want in the main repo.

I actually intend on moving `Variable` here too, when it is completely removed from RxSwift.

I think what @vincent-pradeilles is proposing is sane and welcome here.

freak4pc:
@fpillet I like the idea of `Variable` but FYI I think kruno suggested aliasing it later on in RxSwift core - i guess we'll see what happens :)

jegnux:
> Actually the repo is just about extending the existing abilities of RxSwift itself, with things that don't necessarily fit within the main repo.

Yes, but my point is that it doesnt extend RxSwift. It modifies Swift syntax for closure capturing.

Maybe we could just in introduce a `with(weak:)` operator?

```swift
Observable<Int>
  .interval(1, scheduler:...)
  .take(5)
  .with(weak: self) // -> Observable<(Self, Int)>
  .subscribe(onNext: { strongSelf, i in
      strongSelf.count = I
  })
  .disposed(by: disposeBag)
```

This way, it's a "true" operator and works with every other operators. No need to implement it everywhere.

Implementation suggestion: 
```swift

extension ObservableType {
    func with<T : AnyObject, Out>(weak value: T, resultSelector: (T, Self.E) -> Out) -> Observable<Out> {
        return flatMap { [weak value] element in
            guard let value = value else { return .empty() }
            return .just(resultSelector(value, element))
        }
    }
    func with<T : AnyObject>(weak value: T) -> Observable<(T, Self.E)> {
        return with(weak: value, resultSelector: { value, element in (value, element) })
    }
}
```

freak4pc:
I actually like the idea of making this an operator - that's not a bad idea at all and can save some trouble in implementation. 

For naming - I would go for either 
`unretainined(obj:)` OR `safelyRetain(_ obj:)` but I'm open to other namings. 

I'd want to have @vincent-pradeilles's opinion on this, as he opened this PR. What do you think about this approach? Might save the hassle of going through each of these operators by hand and generalizing this behavior.

jegnux:
I chose `with(weak:resultSelector:)` to match RxSwift `withLatestFrom(_:resultSelector:)`

But this `with()` would do a little bit more: it filters the output if `value` is released.

Actually maybe we should `complete` the subscription when `value` is released.

What about `takeWithAndUntilDeallocation(of:)` ?
```swift
Observable<Int>
  .interval(1, scheduler:...)
  .take(5)
  .takeWithAndUntilDeallocation(of: self) // -> Observable<(Self, Int)>
  .subscribe(onNext: { strongSelf, i in
      strongSelf.count = I
  })
  .disposed(by: disposeBag)
```

freak4pc:
I don't think the method needs to describe that - it also gets out of hand :) 

takeWith isn't proper because the second object isn't an Observable. 

I still side with just `withUnretained(obj:)` for example. The documentation can be explicit about this completing the sequence IMO.

vincent-pradeilles:
@jegnux @freak4pc I like the idea of a `with(weak:)` operator, this makes a lots of sens !

(I'm travelling for my work, so don't expect answers from me before Wednesday ;) )

freak4pc:
Safe travels ! 

I have two problems with the naming 
1. naming it purely with isn't in par with `withLatestFrom`. (e.g. , its' not `with(latestFrom:)`)
2. using `weak` conveys the meaning you're going to get a weak version of the object, which is incorrect. 

`withUnretained(_ object:)` makes most sense, or something around that.

Another option was `wishSafe(_ object:)` but I stand in agreement with @groue that this term is too overloaded with different meanings at this point :)

bobgodwinx:
@jegnux @freak4pc @vincent-pradeilles Yes with `with(weak: obj)` does makes sense but if you think of it critically we already have operatos like `unwrap()` which works fine. Can't we just draw from this style and add something like `weakify()` ? just like the old days :)

freak4pc:
@bobgodwinx I understand where you're coming from - 
The problem is you don't weakify an existing stream of something (which that naming conveys) 

You take the value emitted from an Observable and "add" some safe unretained reference to a difference object _along_ with it - so the `with...` prefix here is pretty crucial in my opinion. 

unwrap() also, is responsible for unwrapping emitted values, and not a "companion" non-Rx object along with the emitted values.

freak4pc:
@bobgodwinx So two more problems with this nameing

1. As I said earlier, `with(weakSelf:)` doesn't follow RxSwift's naming (e.g. it's `withLatestFrom(_:)`, not `with(latestFrom:)`)
2. You can use this method to weakify non-Self objects, not just self. 
3. The naming of "weak" makes people think they will get a weak reference (which is optional) - but that reference is already retained at this point inside the returned value. it's not weak at this point :) So I don't think that naming is also proper. I don't mind using it if you think it conveys more meaning.

So I think it's down between 

```swift
withWeak(_ obj:)

withUneratined(_ obj:)

withSafelyRetained(_ obj:)`
```

Thoughts ?

vincent-pradeilles:
@freak4pc @jegnux how do you want to proceed from now? Do I close my PR and @jegnux creates one of his own with his proposal. Or do I rework my PR by integrating his proposal into it?

freak4pc:
Hey @vincent-pradeilles  - Don't close this for now until we make a final decision on naming. About who opens the final PR, I'm not sure yet :) it doesn't matter that much either way.

bobgodwinx:
ok guys I am fine with the following as proposed by @freak4pc just to be in line with `RxSwift`
```swift
withWeak(_ obj:)
withUnretained(_ obj:)
withSafelyRetained(_ obj:)` 
```

jegnux:
If we had to vote I would +1 on `withUnretained(_ obj:resultSelector:)` (Yes, I do insist with the `resultSelector:` option 😉 )

freak4pc:
The result selector is fine, albeit a bit useless in this specific scenario :) 
As long as it has a default implementation (like the withLatestFrom and combineLatest variations), it's fine by me.

jegnux:
resultSelector is useful when Self.E is already a tuple. It allows to flat `(A, B, C)` instead of `(A, (B, C))`

But yes, we must have a default selector.

freak4pc:
> resultSelector is useful when Self.E is already a tuple. It allows to flat (A, B, C) instead of (A, (B, C))

Fair point! 

Can we move forward with `withUnretained(_obj:resultSelector:)` ? @vincent-pradeilles since you're also traveling, would you be ok with @jegnux preparing the PR for this (if he wants to, at least ;-) ) ?

vincent-pradeilles:
@freak4pc fine with me if he wants to move forward with a PR. Otherwise, I can do it Wednesday evening (french time).

freak4pc:
@jegnux Up to you :) Do you want to take over in a separate PR or do you want to let vincent finalize this?

fpillet:
One note about `withUnretained`, it's a bit weird since Swift uses the `unowned` keyword for this case. Wouldn't it be logical to name it `withUnowned` instead ? For sure it therefore wouldn't match with `withSafelyRetained` anymore...

freak4pc:
withUnowned is also good TBH!

fpillet:
And in the same line (Swift-naming conformance), `withSafelyRetained` should be `withSafelyReferenced`. The `retain`/`release` semantics are completely absent of Swift.

jegnux:
I think `unowned` is confusing. In swift, `unowned` means that the app will crash if the program tries to get the value after it has been released.
It won't be the case here. It will safely filters out emitted value.

That's why I think we should make the subscription `complete` once the given object is released.

fpillet:
Oh okay, hadn't understood your purpose this way. I think that `withWeaklyReferenced` should be the right naming, then (or the short form `withWeak`) since what you're doing is keeping a weak ref.

freak4pc:
@fpillet I don't entirely agree - the reference isn't weak by the time it "gets to you" - you get a "unretained" copy.  The name `withWeak` provides an expectation you're going to get a weak reference (which means it should be Optional, like a [weak] capture group) , but that's not the case so the naming seems confusing to me.

I know you mentioned retainment isn't a thing in Swift but I'd argue the entire point of this feature is avoiding a *retain* cycle - which is a very real pain, in Swift as well :) 

My vote still goes to `withUnretained`

jegnux:
_/me trying again_

```swift
takeWithAndUntilDeallocation(of: self)
```

freak4pc:
Please no :( 

Doesn't work IMO because, 
1. Discoverability is bad
2. Naming is too verbose 

It would be lake naming `map` -> `transformSequenceWithClosure` , IMO

jegnux:
> Discoverability is bad

There are already many `take...` operators so I don't think it would be that bad.

fpillet:
`whileReferenced(_ obj:)` ?

jegnux:
Ok... I have an idea: 
```swift
with(_ obj: T)
```

```swift
Observable<Int>
  .interval(1, scheduler:...)
  .take(5)
  .with(self) // -> Observable<(Self, Int)>
  .subscribe(onNext: { strongSelf, i in
      strongSelf.count = I
  })
  .disposed(by: disposeBag)
```

The fact it's weakly referenced is an implementation detail the caller don't really need to know? Like, we use `delegate` everywhere. Not `weakDelegate`.

fpillet:
The fact that delegates are weak is an implicit convention throughout the OS frameworks (and even then, I occasionally check out the headers to make sure).

`with` amounts to saying `using` and doesn't convey much information about what it's doing.

freak4pc:
That's what you [initially suggested](https://github.com/RxSwiftCommunity/RxSwiftExt/pull/146#issuecomment-376075258) - but I agree with florent it doesn't really convey anything, its a bit too generic. 

With what? a different observable? a value? what's the usage of it? what does it do? 

we can do
`withSafeReference(to:)` 

But I still kinda like `withUnretained(obj:)`

freak4pc:
😂 

how-to-name-things-the-hardest-problem-in-programming-38-638

vincent-pradeilles:
@freak4pc Hello! Do you fell like a consensus on naming the thing has emerged?

freak4pc:
Doesn’t seem like it.
i think it falls between withSafeReference(to:), withUnretained(obj:) and withUnowned(obj:)

Let’s do the last option and we can rename in the review stage

It also needs to have a resultsSelector trailing closure with the default behavior of combining to a tuple:)

freak4pc:
PS Please open it as a new PR so we can have a fresh discussion :) thank you for all your work on this !

PR #149, opened 2018-03-28
Implement withUnretained operator

vincent-pradeilles:
The `withUnowned(_:)` operator provides an unowned, safe to use (i.e. not implicitly unwrapped), reference to an object along with the values of the sequence.

```swift
class Displayer {
    init() { }
    
    deinit {
        print("deinit")
    }
    
    func display(_ value: Any) {
        print(value)
    }
}

let publishSubject = PublishSubject<Int>()
var displayer: Displayer? = Displayer()

if let displayer = displayer {
    _ = publishSubject
        .withUnowned(displayer) // -> Observable<(Displayer, Int)>
        .subscribe(onNext: { displayer, i in
            displayer.display(i)
        })
}

publishSubject.onNext(1)
publishSubject.onNext(2)
publishSubject.onNext(3)
publishSubject.onNext(5)
publishSubject.onNext(8)
displayer = nil // the object referenced by `displayer` gets deallocated
publishSubject.onNext(13)
publishSubject.onNext(21)
publishSubject.onCompleted()
```

```
1
2
3
5
8
deinit
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 master
 -----
+- added `withUnretained(_:)` operator
 
 3.2.0
 -----
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -23,6 +23,7 @@
  
  - [apply()](apply) operator, allows for factoring out repeated Observable chain transformations
  - [unwrap()](unwrap) operator, takes a sequence of optional elements and returns a sequence of non-optional elements, filtering out any `nil` values
+ - [withUnretained(_:)](withUnretained) operator provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object along with the events emitted by the sequence.
  - [ignore(Any...)](ignore) operator, filters out any of the elements passed in parameters
  - [Observable.once(Element)](once) contructor, creates a sequence that delivers an element *once* to the first subscriber then completes. The same sequence will complete immediately without delivering any element to all further subscribers.
  - [mapAt(KeyPath)](mapAt) operator transforms a sequence of elements where each element is mapped to its value at the provided key path
--- a/Playground/RxSwiftExtPlayground.playground/Pages/withUnretained.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/withUnretained.xcplaygroundpage/Contents.swift
@@ -1,0 +1,51 @@
+/*:
+ > # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+ 
+ 1. Make sure you have [Carthage](https://github.com/Carthage/Carthage) installed
+ 1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+ 1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
+ 1. Choose `View > Show Debug Area`
+ */
+
+//: [Previous](@previous)
+
+import RxSwift
+import RxSwiftExt
+
+class Displayer {
+    init() { }
+    
+    deinit {
+        print("deinit")
+    }
+    
+    func display(_ value: Any) {
+        print(value)
+    }
+}
+
+example("withUnretained") {
+    let publishSubject = PublishSubject<Int>()
+    var displayer: Displayer? = Displayer()
+    
+    if let displayer = displayer {
+        _ = publishSubject
+            .withUnretained(displayer) // -> Observable<(Displayer, Int)>
+            .subscribe(onNext: { displayer, i in
+                displayer.display(i)
+            })
+    }
+    
+    publishSubject.onNext(1)
+    publishSubject.onNext(2)
+    publishSubject.onNext(3)
+    publishSubject.onNext(5)
+    publishSubject.onNext(8)
+    displayer = nil // the object referenced by `displayer` gets deallocated
+    publishSubject.onNext(13)
+    publishSubject.onNext(21)
+    publishSubject.onCompleted()
+}
+
+//: [Next](@next)
+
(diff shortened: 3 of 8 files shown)


freak4pc:
I’ll look at this tomorrow - just want to confirm, are you completing the sequence when you can’t retain self ?

vincent-pradeilles:
@freak4pc, yes, `Observable.empty()` returns an empty sequence that completes immediately.

jegnux:
I just looked over Swift and Cocoa APIs:

```swift
struct Unmanaged<Instance : AnyObject> {
  static func passUnretained(_ value: Instance) -> Unmanaged<Instance>
  func takeUnretainedValue() -> Instance
}
```

```swift
class NSValue : NSObject {
  init(nonretainedObject anObject: Any?)
  var nonretainedObjectValue: Any? { get }
}
```

```swift
class NSMapTable<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
  class func weakToStrongObjects() -> NSMapTable<KeyType, ObjectType>
  class func strongToWeakObjects() -> NSMapTable<KeyType, ObjectType>
  class func weakToWeakObjects() -> NSMapTable<KeyType, ObjectType>
}
```

```swift
class NSHashTable< ObjectType : AnyObject > : NSObject {
  class func weakObjects() -> NSHashTable<ObjectType>
}
```

So basically they use `weakObject`, `nonretainedObject` and `unretainedValue`... It's not helping much, unless we want to match closely Swift naming. In that case I think we should use 

#### `withUnretainedValue(_:)`

freak4pc:
You forgot the most common usecase - [unowned self]

PS we shouldn’t use the word value to describe a reference type IMO

These names you found aren’t in reference to the same scenario we’re talkinh about; providing an unowned safe reference to an accompany object (like self which is the main use case) 

Most of these seem unrelated (unmanaged is usually for models like Core Data).

I’m not sure any of these examples support that kind of naming - also do more the different teams in Apple sometimes use different naming conventions for these kinds of things (unfortunately), so it doesn’t always mean the same thing

jegnux:
> You forgot the most common usecase - [unowned self]

That's precisely why I thin `unowned` is not the best choice. An `unowned` reference will lead to a crash if the reference is accessed after deallocation. I'm pretty sure this naming will make developer doubt about its safety.

groue:
@vincent-pradeilles :

>> just want to confirm, are you completing the sequence when you can’t retain self ?
> @freak4pc, yes, `Observable.empty()` returns an empty sequence that completes immediately.

I'm not sure about the best interpretation of @freak4pc's question, but the current implementation, which is based on `flatMap`, has the resulting sequence unable to complete before the base sequence. It won't emit any values after the "unretained" value has been deallocated. But it won't complete either.

@ jegnux:

> That's precisely why I thin `unowned` is not the best choice.

I agree a lot. `unowned` screams "guaranteed crash unless you know what you are doing":

```swift
class Child {
    unowned var parent: Parent // You'd better know what you are doing
}

foo { [unowned self] in
    self.bar() // You'd better know what you are doing
}
```

And we want to avoid any thrill, don't we? "Weak" or "unretained" look safer:

```swift
subject.withUnretained(self).subscribe(onNext: { (strongSelf, value) in ... })
```

freak4pc:
/polls withUnretained(self) withSafelyRetained(self) withWeak(self) withUnowned(self)

freak4pc:
Please see this issue to finalize the naming 
https://github.com/RxSwiftCommunity/RxSwiftExt/issues/150

vincent-pradeilles:
@freak4pc will do the renaming to `withUnretained(_obj:resultsSelector:)` tomorrow ;)

freak4pc:
We have a majority for `withUnretained(_ obj:resultsSelector:)`, so let's go with that.

Also regarding implementation I'll try to review soon :) I'm also confused with when the sequence completes, I'll make sure.

vincent-pradeilles:
@freak4pc renaming has been done 👍

freak4pc:
It’s still missing completion of the outer sequence. To be honest im not sure if this is easily feasible with a flatMap (you’ll have to throw an error).

@jegnux this was your idea IMO - what do you think ?

vincent-pradeilles:
@freak4pc from what I see on the Rx doc, `flatMap` is supposed to complete only when the observable it was called on completes. http://reactivex.io/documentation/operators/flatmap.html

freak4pc:
@vincent-pradeilles That's true but we're designing a new operator here, which isn't specific to flatMap :) flatMap is an implementation detail, in general I think a user might expect the sequence to be terminated if the reference can't be retained. 

Otherwise it means the sequence won't complete but will also not emit _anything_ which is very confusing. 

Any opinions? @RxSwiftCommunity/contributors

jegnux:
Yeah, I think we should complete the sequence if the given object is released. For a simple reason: the sequence will never emit any element ever again. So there's no reason to keep the sequence live.

Indeed it's not possible to do this with a simple flatMap. A combination of something like `.map { throw xxx }` and `.catchError { error in error == xxx ? .empty() : .error(error) }` should work.

freak4pc:
It won't (you could just use `catchErrorJustComplete`) 
So it would discard _all_ errors thrown and not just those specific to this operator. We might need to write something a bit more custom to get around it, possibly.

jegnux:
@freak4pc I'm not sure to understand? 🤔

jegnux:
Why this shouldn't work?

```swift
private enum UnretainedError : Swift.Error {
  case objectReleased
}

public func withUnretained<T : AnyObject, Out>(_ obj: T, resultSelector: @escaping (T, Self.E) -> Out) -> Observable<Out> {
         return map { [weak obj] element -> Out in
             guard let obj = obj else { throw UnretainedError.objectReleased }
             return resultSelector(obj, element)
         }
         .catchError { error -> Observable<Out> in 
             if error == UnretainedError.objectReleased {
                 return .empty()
             } 
             return Observable.error(error)
          }
    }
```

freak4pc:
Uhm, if you make it specific to that error it should be fine. Just need to make tests specific to completion etc.

vincent-pradeilles:
@jegnux @freak4pc duly noted, I'll make the tests and modification later today ;)

freak4pc:
Sounds good. Thanks @vincent-pradeilles ! We're very close here :)

vincent-pradeilles:
@freak4pc @jegnux  done 👌

freak4pc:
Hey vincent, thanks! Have some notes, will accumulate and send soon.

freak4pc:
Ok this is ready to go as far as I'm concerned! 
Would appreciate any notes if anyone wants to make a review of the revised code. 

@vincent-pradeilles @jegnux @RxSwiftCommunity/contributors

freak4pc:
:shipit:

PR #153, opened 2018-04-10
Add fractionComplete binder on UIViewPropertyAnimator reactive extension

andreanmasiro:
With the recent `UIViewPropertyAnimator` reactive extension, I had the idea to add fraction complete binder to it, as I have already used it in a reactive-like way, but with no Rx.

I found a crash when using the `rx.fractionComplete` and then calling `rx.animate()`, I will debug it sometime and try to remove it before the merge.

I'd like to hear your opinions, so I can make it good and then prepare the Readme file and the demo in the playground.


freak4pc:
That's interesting in general IMO. Would you say its fitting only as a binder or as a ControlProperty ?

andreanmasiro:
I have done a little research and have not found a way to observe the changes to it. It seems that KVO does not updates the value until the animation is complete, maybe Apple will find a need to add some kind of observing to this sometime in the future.

freak4pc:
Thanks for researching ! I’m basically OK with this but could you suggest a use case so I can truly make sense of adding we a Binder as part of the library ?

andreanmasiro:
Sure. The common use case is the need to control the animation state of some property by reacting to some other property. 

For example: I worked on a blur animation on the navigation bar that was controlled by my table view content offset (just like **twitter** profile). What I did then was to observe the offset with the scroll view delegate and manually update the `fractionComplete` in my animator. By using this binder and other RxCocoa features, I could simply get the content offset reactively and bind it to my animator.

freak4pc:
Yup that makes total sense to me. If you want to land this PR, You'll have to make it complete - eg. tests, playground page, readme - everything in the Contribution guide. Let me know if you need any help!

andreanmasiro:
Working on it. Should I change the version to 3.4.0 or just 3.3.1, since the last change was also related to `UIViewPropertyAnimator`?

CHANGELOG.md
```diff
@@ -5,9 +5,13 @@ master
 -----
 - added `withUnretained(_:)` operator
 
+3.4.0
+-----
+- added UIViewPropertyAnimator `fractionComplete` reactive binder
```
freak4pc: This should be under master (e.g. above the `withUnretained` addition)

Issue #93, opened 2017-06-19
Adding mapElements operator


TheAdamBorek:
## Adding `mapElements` operator

In every project where I use RxSwift I always end up with adding small helper for mapping elements inside the array in `Observable<[Type]>`. Maybe it's time to include it into one library with common, additional Rx operators ;)

### Motivation for inclusion

My motivation is simple. This... :
```swift
theObservable
  .mapElements(TheModel.init)
```

is more readable than... :

```swift
theObservable
 .map { array: in 
   return array.map(TheModel.init)
}
```

### Example of use
Example can be found just above ;)

mosamer:
I may also suggest adding `filterElements` operator

jdisho:
What do you think about this @freak4pc ?

freak4pc:
Totally missed this one, thanks for the bump @jdisho. I think this could be a nice addition potentially, just want to make sure I got you correctly.

Are you looking for something like this @TheAdamBorek : 

```swift
extension ObservableType where E: Collection {
  func mapMany<R>(_ transform: @escaping (E.Element) throws -> R) -> Observable<[R]> {
    return map { collection in
      try collection.map(transform)
    }
  }
}
```

Usage: 
```swift
struct Thing: CustomStringConvertible {
  let number: Int
  init(number: Int) {
    self.number = number
  }

  var description: String { return "#\(number)" }
}

Observable.of([1, 2, 3, 4])
  .mapMany(Thing.init)
```

Output:
`-> Event next([#1, #2, #3, #4])`

jdisho:
Why calling it `mapMany` @freak4pc? I think `mapElements` is more clear.

freak4pc:
I'm personally not a fan of `mapElements`, since the original operator isn't called `MapElement` in singular... 

When you map emissions coming from an observable stream you are also `mapping elements` so I don't think this creates a proper "plural" distinction (e.g. this works only on `Collection` types)

We can put it up for a vote of names like we did in #150

TheAdamBorek:
To be honest, I also like `mapMany` more than `mapElementes` (since Observable already has elements) ;)

However, the PR is quite old. I assume I would have to update it, right?

TheAdamBorek:
@jdisho sure, go ahead ;

freak4pc:
Let's continue discussion over at #154 :)

PR #154, opened 2018-05-06
mapMany Operator

jdisho:
For the proposed operator #93 🎉

### DIFF
--- a/Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
@@ -1,0 +1,32 @@
+/*:
+ > # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+
+ 1. Make sure you have [Carthage](https://github.com/Carthage/Carthage) installed
+ 1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+ 1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
+ 1. Choose `View > Show Debug Area`
+ */
+
+//: [Previous](@previous)
+import RxSwift
+import RxSwiftExt
+
+example("mapMany") {
+    let ints = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+    let intsStream = Observable.just(ints)
+
+    let strings = ["Make, sure, Carthage, RxSwiftExtPlayground"]
+    let stringsStream = Observable.just(strings)
+
+    intsStream.mapMany { $0 * $0 }
+        .subscribe(onNext: { result in
+            print(result)
+        })
+
+    stringsStream.mapMany { $0.lowercased() }
+        .subscribe(onNext: { result in
+            print(result)
+        })
+}
+
+//: [Next](@next)
--- a/Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
@@ -11,37 +11,35 @@
 import RxSwift
 import RxSwiftExt
 
-struct SomeModel: CustomStringConvertible {
-    let number: Int
-    var description: String { return "#\(number)" }
-
-    init(_ number: Int) {
-        self.number = number
-    }
-}
-
 example("mapMany") {
-    let ints = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
-    let intsStream = Observable.just(ints)
+    let numbers = Observable.of(1...10)
+    let strings = Observable.of(["RxSwift", "is" ,"awesome", "along", "with", "RxSwiftCommunity"])
 
-    let strings = ["Make, sure, Carthage, RxSwiftExtPlayground"]
-    let stringsStream = Observable.just(strings)
-
-    intsStream.mapMany { $0 * $0 }
+    // Map many using a model initializer
+    numbers.mapMany(SomeModel.init)
         .subscribe(onNext: { result in
             print(result)
         })
 
-    intsStream.mapMany(SomeModel.init)
+    // Map many with a transformation closure
+    numbers.mapMany { $0 * $0 }
         .subscribe(onNext: { result in
             print(result)
         })
 
-
-    stringsStream.mapMany { $0.lowercased() }
+    strings.mapMany { $0.lowercased() }
         .subscribe(onNext: { result in
             print(result)
         })
+
+    struct SomeModel: CustomStringConvertible {
+        let number: Int
+        var description: String { return "#\(number)" }
+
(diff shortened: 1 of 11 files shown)


Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    let strings = ["Make, sure, Carthage, RxSwiftExtPlayground"]
+    let stringsStream = Observable.just(strings)
+
+    intsStream.mapMany { $0 * $0 }
+        .subscribe(onNext: { result in
+            print(result)
+        })
+
+    stringsStream.mapMany { $0.lowercased() }
```
freak4pc: Can we add an example with a fake initializer (e.g. add a fake struct and initialize many of it with `mapMany`) ?

Source/RxSwift/mapMany.swift
```diff
@@ -0,0 +1,24 @@
...
+
+extension ObservableType where E: Collection {
+
+    /**
+     Projects elements of each element of an observable collection into a new form.
+     - parameter transform: A transform function to apply to each source element.
+     - returns: An observable collection whose elements are the result of invoking the transform function on each element of source.
+     */
+
+    public func mapMany<T>(_ transform: @escaping (E.Element) -> T) -> Observable<[T]> {
```
freak4pc: I think the `transform` block needs to throw like `map` itself: https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Observables/Map.swift#L20

What about the implementation example I put in https://github.com/RxSwiftCommunity/RxSwiftExt/issues/93#issuecomment-378631505 ?

Tests/RxSwift/MapManyTests.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Created by Joan Disho on 06.05.18.
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import XCTest
+import RxSwift
+import RxTest
+
+class MapManyTests: XCTestCase {
+    fileprivate func runAndObserve<C: Collection>(_ sequence: Observable<C>) -> TestableObserver<C> {
```
freak4pc: Same here like the sample project - please add something that tests mass initialization of a data model, like a struct. 

Thanks !

freak4pc:
P.S @jdisho If you'll have time to do this in the next few days we can push it in the upcoming release :) Don't mind waiting since its a few minor fixes.

Source/RxSwift/mapMany.swift
jdisho: It is a bit unclear why it should throw 🤔

freak4pc: Because the original map method accepts a throwing closure, so we should accept a throwing closure as well:
func map<T>(_ transform: (Self.Element) **throws** -> T) rethrows -> [T]

Source/RxSwift/mapMany.swift
```diff
@@ -16,7 +16,7 @@ extension ObservableType where E: Collection {
      - returns: An observable collection whose elements are the result of invoking the transform function on each element of source.
      */
 
-    public func mapMany<T>(_ transform: @escaping (E.Element) throws -> T) -> Observable<[T]> {
+    public func mapMany<T>(_ transform: @escaping (E.Element) throws -> T) rethrows -> Observable<[T]> {
```
freak4pc: There's no need for a regrow retrow IMO.
Look at the original signature from RxSwift map:

e7bbb63d-ca2e-4f8e-b984-df05005af0b1

jdisho: Don't mind reverting the commit, but a func can be a rethrowing func to show that it throws an error only if it’s function parameter throws an error. 🤔

freak4pc: Uhm if that would be the case than the map throws without rethrow doesn’t make sense ? I think in general we’d want it to behave much like a regular map so I think It’s best to keep the signature similar

Tests/RxSwift/MapManyTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    let number: Int
+    var description: String { return "#\(number)" }
+
+    init(_ number: Int) {
+        self.number = number
+    }
+}
+
+class MapManyTests: XCTestCase {
+    fileprivate func runAndObserve<C: Collection>(_ sequence: Observable<C>) -> TestableObserver<C> {
```
freak4pc: can you get rid of `fileprivate` here (and also on top) ? 
It's irrelevant given the fact its part of a test suite, and the method is part of the class itself.

Playground/RxSwiftExtPlayground.playground/Pages/mapMany.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,47 @@
...
+ 1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+ 1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
+ 1. Choose `View > Show Debug Area`
+ */
+
+//: [Previous](@previous)
+import RxSwift
+import RxSwiftExt
+
+fileprivate struct SomeModel: CustomStringConvertible {
```
freak4pc: Same here, no reason for ACL at all given the fact its constrained to the playground page

freak4pc:
@jdisho Did a bit of cleanup, let me know what you think :) 
I'll do another pass of review tomorrow. 

Also noticed some issue with `toSortedArray` playground page and missing CHANGELOG entry.

freak4pc:
There seems to be a few missing operators in the playground page and README 🤔 not sure why. Need to find the time to look into it.

jdisho:
Can this operator be merged anytime soon @freak4pc ? 😊 

freak4pc:
I'm gonna merge this as soon as CI passes. 

And yeah that'd be great but to be honest i feel like we're going to have to add some automation to make sure documentation is synced across README -> Playgrounds -> Comments etc. Might be some work but we'll see later on :)

freak4pc:
Something with your tests seem wrong, can you look into that? 

5baeac03-cf2c-4959-8ec1-e0a5b3e3cc55

jdisho:
Strange, CI is successfully passed.😕

freak4pc:
Yeah, the fact CI passes is a bug that you should ignore - but the build clearly fails

freak4pc:
Ok, so CI is fixed - now time to fix your tests :) @jdisho

jdisho:
I don't understand, the test passed in Xcode 😕

freak4pc:
@jdisho Tried bumping CI Xcode to 9.4.0, lets see if that helps

jdisho:
Apparently no 
screen shot 2018-07-08 at 17 38 02

freak4pc:
Here we go :) @jdisho

freak4pc:
@jdisho Usually the author of the PR should not  be the one to merge it, just FYI - but no biggie :)

PR #155, opened 2018-05-12
Enumerated operator

jdisho:
In this PR I am introducing the `enumerated` operator. This operator converts an observable into another observable that also emits the index count of the emission. 🎉


freak4pc:
Hey @jdisho - I think this already exists in core, let me know if you think differently : 

https://github.com/ReactiveX/RxSwift/blob/f1e215a3fe1303568e415095cb00a6ca2a27b706/RxSwift/Observables/Enumerated.swift#L18

jdisho:
Ahh alright! I didn't know that enumerated  exists in core. 😊

PR #157, opened 2018-05-30
update .gitignore

iDurian:
git ignore Carthage/ folder


freak4pc:
Seems to already be on master, but thanks for your contribution!

PR #159, opened 2018-06-15
Make pausableBuffered robust to reentrancy

matthewvalentine:
pausableBuffered currently suffers from some issues with reentrancy.

1. If you trigger a flush while one is already happening, you'll end up with a duplicate flush happening in the middle of the first one. Among other ways, this can happen if the pauser is reentrant.
2. If you send in another value while a flush is happening, it will be emitted immediately, disrupting the original order of events. Among other ways, this can happen if the underlying observable is reentrant.

This pull request fixes this by:

1. Instead of each flush restarting from the beginning, the flushes share and increment a collective flush index.
2. The pauser is filtered through distinctUntilChanged, and the pause status is only updated _after_ the any triggered flush is completed. This way, events that come in during a flush find the buffer in a "paused" state and add themselves to the end of the buffer. Since the flush is being done with an index, this presents no issue - it just ends up as if it were part of the flush.

I don't know that this is the _best_ way to solve the issue but it seems to work. Something deficient with this solution is that an infinitely re-entrant underlying observable would grow the buffer to infinite size.

### DIFF
--- a/Source/RxSwift/pausableBuffered.swift
+++ b/Source/RxSwift/pausableBuffered.swift
@@ -35,24 +35,29 @@ extension ObservableType {
             }
 
             var paused = true
+            var flushIndex = 0
             let lock = NSRecursiveLock()
 
             let flush = {
-                for value in buffer {
-                    observer.onNext(value)
+                while flushIndex < buffer.count {
+                    flushIndex += 1
+                    observer.onNext(buffer[flushIndex - 1])
                 }
-                buffer.removeAll(keepingCapacity: limit != nil)
+                if buffer.count > 0 {
+                    flushIndex = 0
+                    buffer.removeAll(keepingCapacity: limit != nil)
+                }
             }
 
-            let boundaryDisposable = pauser.subscribe { event in
+            let boundaryDisposable = pauser.distinctUntilChanged(==).subscribe { event in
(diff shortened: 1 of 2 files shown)


freak4pc:
This looks good in general (didn't go through the whole code, yet) 

I have once concern we should consider:

If there are some older code-bases using the existing version of `pausableBuffered`, they might be counting on this existing (semi-broken) behavior, to a point where "fixing" this will break their code. I would like to get more feedback from other community members regarding this. But I think this might be safe. (@fpillet if you have a minute, that would be great. Also, the original author of this operator is @tgyhlsb but not sure if he's still an active community member at this point).

Thanks for your contribution!
Shai.

matthewvalentine:
I can certainly see that concern. Especially since AFAIK re-entrancy is already undefined or pathological behavior in other situations. So even the current form probably already fulfills its contract - this operator is just more broken than most because it can result in replay rather than just reorder.

freak4pc:
OK, thought of this some more. I think it's fine if we bump the minor version here. The existing tests weren't altered in any way, so they still describe how this operator works even after this fix. For that case, I'm probably OK with merging it but I would really like some more eyes on this @RxSwiftCommunity/contributors. 

@twittemb You use pasuableBuffered quite a lot IIRC correctly. could you share your thoughts?

PR #166, opened 2018-07-16
Bind observable to collection of observers

### LINKED ISSUE #163: Bind observable to collection of observers
### Name and description

An extension on ObservableType to allow `.bind(to:)` with a collection of observers, and an extension on Collection (where Iterator.Element == Disposable) to allow `.disposed(by:)` with a collection of Disposables.

### Motivation for inclusion

Useful when you have a single observable binding to a lot of observers. A big use case is when implementing forms and having a single observable bind to isEnabled on a multitude of controls.

### Example of use

```swift
isEnabledSubject
  .bind(to: myCollectionOfTextFields.map({$0.rx.isEnabled}))
  .disposed(by: disposeBag)
```

Implementation here: https://gist.github.com/sgtsquiggs/0b1dc9dbd6fe51f107d4082ac7c448cd

sgtsquiggs:
An extension on ObservableType to allow `.bind(to:)` with a collection of observers, and an extension on Collection (where Iterator.Element == Disposable) to allow `.disposed(by:)` with a collection of Disposables.

Fixes #163


bobgodwinx:
@sgtsquiggs This is a nice feature. I really like and already have something exactly like this in my codebase. It would be nice to add `drive(in: ...)` also what do you guys think? @freak4pc

freak4pc:
@bobgodwinx There was discussion in #163 where I also agreed it was a pretty great addition. 

For `drive(in:)` I see way fewer use cases... Any examples?

sgtsquiggs:
I guess if you're manipulating UI controls (as in the example in #163) you may already be using Drivers to do so.

freak4pc:
In that case just `drive(...)` / `drive([..])`, IMO

bobgodwinx:
@freak4pc it's useful when you have more than one UI that changes on an event. So you just get the updates and `drive(in: ...)` at once.  Another thing is discoverability while coding that's why I proposed `drive(in: ..)` so to make it stand out from the rest of the available methods. But `drive(...)` is also fine.

freak4pc:
I thin that if we're overloading `bind(to:)` with Collections, we should do the same for `drive()`... `drive(in:)` also isn't intuitive IMO

bobgodwinx:
Sounds like a deal let's go with `drive(...)` 👍🏿

freak4pc:
Cool. So, @sgtsquiggs would you like to take care of that? 

Also, please look at the contribution guidelines: https://github.com/RxSwiftCommunity/RxSwiftExt/blob/master/CONTRIBUTING.md#submitting-a-pull-request

You need to add a playground page and update the README documenting these operators.

sgtsquiggs:
Hmm I can't seem to get the playground working. I'm not too familiar with Carthage, so may be that. It seems to be loading latest released RxSwiftExt from this repo instead of my local modified copy - so it does not have my new operator available. I had the same issue with adding tests; my test is the only one that is using `@testable import RxSwiftExt`.

Any pointers?

freak4pc:
@sgtsquiggs If you want to put your work on this branch, I can try and take a look

freak4pc:
@sgtsquiggs You forgot to make your extensions `public` ;-) 
I also added a variadic option. Feel free to pull and proceed from here.

sgtsquiggs:
I seem to have broken the Playground with this last commit. It hangs before executing any code.

Readme.md
```diff
@@ -556,6 +557,28 @@ next((Test Class, 13))
 completed
 ```
 
+#### bindCollection
```
freak4pc: This should be bindMany like the other places.

Tests/RxCocoa/DriveManyTests+RxCocoa.swift
```diff
@@ -0,0 +1,43 @@
...
+
+class DriveManyTests: XCTestCase {
+    func testDriveMany() {
+        let scheduler = TestScheduler(initialClock: 0, simulateProcessingDelay: false)
+
+        SharingScheduler.mock(scheduler: scheduler) {
+
+            let values = ["one", "two", "three", "four"]
+
+            let observers = [scheduler.createObserver(String.self),
```
freak4pc: Can you make a test that confirms that All disposables are disposed ?

sgtsquiggs: Can do. I did not test because of CompositeDisposable's existing tests.

sgtsquiggs:
Does this PR need any additional work?

freak4pc:
@sgtsquiggs Doesn't seem so. I want to give the discussion in ReactiveX/RxSwift#1702 a fair opportunity. Let's give it two more weeks or so, things move a bit slower there :)

sgtsquiggs:
In light of https://github.com/ReactiveX/RxSwift/pull/1702#issuecomment-412154549 I think it would be best to use `publish` and `connect` instead of `share`.

freak4pc:
@sgtsquiggs There has been good movement on ReactiveX/RxSwift#1702 . Kruno suggests adding this only in RxSwift 5, but im not 100% sure since it doesn't introduce breakage. But he's open to adding it, so I suggest we keep following on Core.

freak4pc:
This is coming in RxSwift 5. Closing. https://github.com/ReactiveX/RxSwift/pull/1702

PR #167, opened 2018-07-19
Update RxSwiftExt.podspec

xiaodoudou:
Hey guys,

Did add curry to swift3 branch.

Cheers,


freak4pc:
@xiaodoudou This won't help you much unless you'll link directly to the swift3 branch from your Podfile, would that work?

freak4pc:
I have no good way to make sure this branch actually works well at this point, but if it solves the problem for you, we'll merge this for now. If there's any issue we can just revert, but this seems quite additive anyways.

PR #171, opened 2018-08-06
Adds non-polling buffer.

BrianSemiglia:
Projects each element of an observable sequence into a buffer that's sent out when a given amount of time has elapsed since the first event after the last flush. This is similar to the standard buffer except that events are not produced as the result of a polling timer which provides performance improvements when a shorter `timeSpan` is specified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 master
 -----
+- added `nonPollingBuffer` operator
 - added `mapMany` operator
 - added `toSortedArray` operator
 
--- a/Playground/RxSwiftExtPlayground.playground/Pages/nonPollingBuffer.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/nonPollingBuffer.xcplaygroundpage/Contents.swift
@@ -1,0 +1,42 @@
+/*:
+> # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+
+1. Make sure you have [Carthage](https://github.com/Carthage/Carthage) installed
+1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
+1. Choose `View > Show Debug Area`
+*/
+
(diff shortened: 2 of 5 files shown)


freak4pc:
Hey @BrianSemiglia ! Really cool contribution :) I'll dig into this a bit later and invite @RxSwiftCommunity/contributors to do the same. 

I can already say I have an issue with the name, because "nonPollingBuffer" doesn't sound like a method name, but sounds like a strategy name. 

Perhaps we can have our own `buffer(strategy: BufferingStrategy[default/nonPolling])`, default just relaying to the Core representation. 

What do you think?

BrianSemiglia:
I agree that the label could use some work seeing as it specifies what it doesn't do. I'm not so sure about serving as an abstracted dispatch to the native buffer though; my preference would be to have implementations call the desired buffer directly as it would 1.) look familiar to an audience familiar with RxSwift and 2.) prevent needing to update the function interface due to changes to the native buffer.

Would a label like `lazyBuffer` or `passiveBuffer` suffer from the same problem of specifying a strategy?

freak4pc:
I think the prefix has to be `buffer`, e.g `bufferLazily`, `bufferPassively`. 

I'm not suggesting to abstract to the native buffer, I suggest that only for the `default` option. 

e.g. 

```swift
enum BufferStrategy {
    case `default`
    case nonPolling
}

func buffer(with strategy: BufferStrategy) {
    switch strategy {
    case .default: // dispatch to core implementation
    case .nonPolling: // your implementation
    }
}

**Comment by BrianSemiglia** [comment]
My concern with the dispatch approach was the need to update `buffer(with:)` when changes to the signature of the core buffer method were made. It's not the worst trade-off though.

If the dispatch approach is the way to go, I'd suggest a label other than `default` only because it doesn't describe its strategy but rather the method's preference for it. Maybe something more like:

```
enum BufferingStrategy {
    case active
    case passive
}
```

freak4pc:
I think `default` (or `standard`) is good to not confused people with the standard behavior... I think active/passive might be confusing, don't feel too religious about this though. 

@RxSwiftCommunity/contributors Any thoughts on naming for this operator, strategies, opinions etc?

Issue #122, opened 2017-12-09
UIScrollView `reachedBottom`


mosamer:
### Name and description
`UIScrollView` reactive extension for reaching bottom while scrolling.

### Motivation for inclusion
While `UIScrollView` already contains reactive extensions for `contentOffset` and scrolling to top, there is no such extension for reaching bottom of the scrollable content. This is commonly used to trigger pagination requests.

### Example of use
```
scrollView.rx
   .reachedBottom(edgeOffset: 10.0)
   .bind(to: viewModel.loadNextPage)
   .disposed(by: bag)
```

freak4pc:
I agree this would be a great addition. 
Here's a simple implementation from Kruno on RxFeedback that we might want to use: 

https://github.com/NoTests/RxFeedback.swift/blob/master/Examples/GithubPaginatedSearch.swift#L192

I would just consider two things here:
1. the naming.  
2. does RxSwiftExt even have any support for "UI"-level extensions at this point?

freak4pc:
Hey @mosamer - This is in my to-do list to add in RxSwiftExt fairly soon. Closing this for now :)

freak4pc:
@mosamer Are you still interested in going after this implementation? :)

mosamer:
@freak4pc I already had an implementation when posted this first time, let me find it and create a PR, hopefully, this weekend.

jdisho:
Hey guys, what do you think about this implementation? 
I can take over if @mosamer is okay with it 😊 

```swift 
extension Reactive where Base: UIScrollView {
    func reachedBottom(offset: Double = 0.0) -> ControlEvent<Bool> {
        let observable = contentOffset
            .flatMap { [weak base] contentOffset -> Observable<Bool> in
                guard let scrollView = base else { return .empty() }
                let visibleHeight = scrollView.frame.height
                    - scrollView.contentInset.top
                    - scrollView.contentInset.bottom
                let y = contentOffset.y + scrollView.contentInset.top
                let threshold = max(offset, scrollView.contentSize.height - visibleHeight)

                return .just(y > threshold)
        }

        return ControlEvent(events: observable)
    }
}
```

fpillet:
@jdisho why go for `flatMap` where a simple `map` would have done the job ? If the `guard` statement fails, just return false. Since you're not doing anything special on this error, I don't think using the more expensive `flatMap` is worth it.

Other than this, 👍

fpillet:
[...]

PR #173, opened 2018-08-17
Implement UIScrollView.reachedBottom

jdisho:
In this PR I implement the `reachedBottom` operator. #122


Source/RxCocoa/reachedBottom.swift
```diff
@@ -0,0 +1,36 @@
...
+
+extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the the bottom of the UIScrollView is reached.
+
+     - parameter offset: A threshhold indicating the bottom.
+     - returns: A boolean value showing if the bottom is reached or not.
+     */
+    
+    func reachedBottom(_ offset: CGFloat = 0.0) -> ControlEvent<Bool> {
```
freak4pc: I'm not sure if we should use a ControlEvent or a Signal in this case. It's not really a control event (e.g. one that stems from a UIControl.Event).

Also, IRT the implementation - it mostly looks O.K but missing all the tests, playground page, etc.

MortyMerr: Hey @freak4pc `Signal` looks fine for you? I am ready to provide tests & playground page if current implementation is ok

MortyMerr: I think `.reachedBottom` is the most popular boilerplate thing which have to be part of this lib

jdisho: @MortyMerr I think observable would me more suitable. Please feel free to change it and add some tests. 😊

MortyMerr: @jdisho why not Signal?
Working on it right now, ready to discuss

MortyMerr:
Here is PR with tests&playground - https://github.com/RxSwiftCommunity/RxSwiftExt/pull/207

jdisho:
Why did you create another PR? You can add you commits here.

freak4pc:
Hey @MortyMerr - I do agree with Joan here. 
As much as the intent is awesome (for you to contribute) - the original PR and idea was brought here by @jdisho, it would be better to overlay your changes on top of this branch if possible. 

Thanks !

Source/RxCocoa/reachedBottom.swift
freak4pc: I think it should definitely be a `Signal` or `ControlEvent` as they both represent UI events. These traits also provide a main-thread guarantee (which you'd want for a UI Element).

We could use `Observable` but that's more generic and not "specific". I think `Signal` makes most sense.

freak4pc: Hey @MortyMerr :) 

OSS is an open discussion, we're not commercially developing this project :)

I'm not a Dictator and don't possess ultimate knowledge; I experiment, and sometimes change my mind and I think that's OK ;-) Some more patience would be appreciated. 

My initial instinct and current is to use Signal because of the guarantees.

MortyMerr:
I did not mean u are dictator, sorry if it looks so 😌
Ok, will do

jdisho:
Actually, I agree with  @freak4pc . Better to be a `Signal`. This is a good case when we are interested in one element, rather than a stream of elements

MortyMerr:
But Signal is a stream of elements.

freak4pc:
`Signal` is identical to `Driver`, it just doesn't replay 1. It's a stream.

MortyMerr:
Yes, i know)

Source/RxCocoa/reachedBottom.swift
bobgodwinx: Hi everyone just to give my opinion here I have one question: 

- When you tap a button what is it? = `ControlEvent`
- When you select a cell what is it? = `ControlEvent`
- When `didScroll` what is it? = `ControlEvent`

So this should be a `ControlEvent` IMHO because you a scrolling and you reached the `Bottom`of the page due to your scrolling actions which was a `ControlEvent`. However `Signal` or `Driver` is also fine, but `ControlEvent` fits better.

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,28 @@
...
+import RxCocoa
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the the bottom of the UIScrollView is reached.
+     - parameter offset: A threshhold indicating the bottom.
+     - returns: Sequence that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> Signal<Void> {
+        return contentOffset.filterMap { contentOffset in
```
jdisho: I don't see the need for `filterMap`. Why not `map`?

jdisho: don't forget `[weak base]`

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Copyright © 2019 RxSwift Community. All rights reserved.
+//
+
+import UIKit
+import RxSwift
+import RxCocoa
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the the bottom of the UIScrollView is reached.
```
jdisho: nit: remove one _the_

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,28 @@
...
+     - parameter offset: A threshhold indicating the bottom.
+     - returns: Sequence that emits when the bottom of the base UIScrollView is reached.
```
jdisho: suggestion: `Single` that emits when the bottom of the `UIScrollView` is reached.

Source/RxCocoa/reachedBottom.swift
MortyMerr: @bobgodwinx I have nothing to argue against this. I also prefer `ControlEvent`. @freak4pc what do u think?

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+import UIKit
+import RxSwift
+import RxCocoa
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the the bottom of the UIScrollView is reached.
+     - parameter offset: A threshhold indicating the bottom.
```
jdisho: suggestion:  A threshold indicating the bottom of the `UIScrollView`.

Source/RxCocoa/UIScrollView+reachedBottom.swift
jdisho: suggestion: Emits if the bottom of the `UIScrollView` is reached.

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Because we should filter events in case contentOffset is not near bottom

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Fixed

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: It's not a Single

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Fixed

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Fixed

Source/RxCocoa/UIScrollView+reachedBottom.swift
jdisho: Damn! Totally overlooked it. I thought we where talking about Singles.

Source/RxCocoa/reachedBottom.swift
freak4pc: OK, after looking into Core source code I think we can go with `ControlEvent` as well. 

FYI though, note that most of these (if not all) were added before Signal was introduced. So they don't all necessarily fit a ControlEvent. Anyways, sure, let's go with it.

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: There is no need in it, no reference cycle here

freak4pc: Not positive you're right, there might be a transitive cycle with `contentOffset`'s implementaion and `DelegateProxy`. Regardless, I'd never trust a "3rd party" implementation to provide retain-cycle safety for me. Unless I'm capturing the objective myself, the inner implementation could possibly "change under my feet". I'd add the capture group personally.

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,28 @@
...
+import RxCocoa
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the bottom of the UIScrollView is reached.
+     - parameter offset: A threshhold indicating the bottom of the UIScrollView.
+     - returns: ControlEvent that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
+        let source = contentOffset.filterMap { contentOffset -> FilterMap<Void> in
```
jdisho: I think it would be better if we are not using custom operators in custom operators. 
I would like to avoid this. My suggestion would be

```swift 
func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
        let observable = contentOffset
            .map { [weak base] contentOffset -> Bool in
                guard let scrollView = base else { return false }
                let visibleHeight = scrollView.frame.height
                    - scrollView.contentInset.top
                    - scrollView.contentInset.bottom
                let y = contentOffset.y + scrollView.contentInset.top
                let threshold = max(offset, scrollView.contentSize.height - visibleHeight)

                return y > threshold
            }
            .distinctUntilChanged()
            .skipWhile { $0 == true }
            .map { _ in }

        return ControlEvent(events: observable)
```

MortyMerr: The argument about atomic operators is not obvious for me. There is a library of operators, why can't we u use them in implementation on new operators?
Thank you for your implementation, but I know how to replace `filterMap`. Two lines of code. However, I can't see any pros in these code duplication.
BTW, `[weak self]` is not needed, there is no reference cycle here. I check it with debugging mode (`RxSwiftResources.Count`)

MortyMerr: > There are cases that I wouldn't like to import the whole library, but just grab the part that I am interested. 
You can't do it with `fromAsync` operator for instance.
If there is a requirement "Don't use other operators to implement new one" it should be discussed and placed to CONTRIBUTING.md

freak4pc: I think if we're talking about two lines of code, It's better to not reuse internal operators that aren't part of core. Many people copy and paste operators from the source code since they don't want to take in a full library (another example of this is https://github.com/SwifterSwift/SwifterSwift)

freak4pc: In general - I think that unless it's a necessity I'd rather not couple RxSwiftExt operators with other RxSwiftExt operators

jdisho: Where do you see the code duplication?

I gave one argument: 

> There are cases that I/others wouldn't like to import the whole library, but just grab the part that I/they care.

freak4pc: > If there is a requirement "Don't use other operators to implement new one" it should be discussed and placed to CONTRIBUTING.md

I don't mind adding it to CONTRIBUTING.md. Unless it's crucial I don't see a reason to reuse these internal operators.

MortyMerr: Ok, let's add it there please. Could i add "You should not use operators from RxSwiftExt to implement new operators" in these PR or it's better to create another?

freak4pc: I'll take care of it.

MortyMerr: Tnank you :)

MortyMerr: Replaced `.filterMap` on `.filter` + `.map`

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,30 @@
...
+     - returns: ControlEvent that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
+        let source = contentOffset.map { contentOffset in
+            let visibleHeight = self.base.frame.height - self.base.contentInset.top - self.base.contentInset.bottom
+            let y = contentOffset.y + self.base.contentInset.top
+            let threshold = max(offset, self.base.contentSize.height - visibleHeight)
+            return y >= threshold
+        }
+        .filter { $0 }
```
freak4pc: By the way, not sure if you'd want to but you can replace most of the implementation with a flatMap:

```swift
let source = contentOffset.flatMap { contentOffset -> Observable<Void> in
            let visibleHeight = self.base.frame.height - self.base.contentInset.top - self.base.contentInset.bottom
            let y = contentOffset.y + self.base.contentInset.top
            let threshold = max(offset, self.base.contentSize.height - visibleHeight)
            
            guard y >= threshold else { return .never() }
            return .just(())
        }
```

You don't have to but it's just a suggestion if you want to avoid passing through 3 operators (should be negligible).

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,53 @@
...
+import RxSwift
+import RxCocoa
+import RxSwiftExt
+import PlaygroundSupport
+import UIKit
+
+/*:
+ ## reachedBottom
+
+ The `reachedBottom` operator provides a sequence that emits every time scrollview is scrolled to the bottom.
```
freak4pc: This should be either
```suggestion
 The `reachedBottom` operator provides a sequence that emits every time the scroll view is scrolled to the bottom.
```

Or: 

```suggestion
 The `reachedBottom` operator provides a sequence that emits every time the `UIScrollView` is scrolled to the bottom.
```

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: I saw this option of the implementation. I think it's better to use `map` & `filter` composition here, because it's not a common way of using `flatMap`.

>most of the implementation

I can see one line saving. Not sure it's really big improvement. What do you think?

freak4pc: Matter of preference, I don't care at all :) 
There's no problem with using filter + map and there's nothing wrong with using `flatMap` for this as well. Feel free to disregard, I just offered it as an option.

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,30 @@
...
+     - parameter offset: A threshhold indicating the bottom of the UIScrollView.
+     - returns: ControlEvent that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
+        let source = contentOffset.map { contentOffset in
+            let visibleHeight = self.base.frame.height - self.base.contentInset.top - self.base.contentInset.bottom
+            let y = contentOffset.y + self.base.contentInset.top
+            let threshold = max(offset, self.base.contentSize.height - visibleHeight)
+            return y >= threshold
+        }
```
jdisho: You would need `distinctUntilChanged` here to remove duplicate states.

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Thank you) Good option, but i am ok with current impl

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: I don't want to remove dublicate states, because scrollView can reach bottom more than once

freak4pc: Don't think so - you want to know when it arrives, not the 50 times you'd cross the boundary, right?

Once you reach the bottom you can't "reach the bottom again" until you go up. Did you try this when you say it causes strange behavior ?

MortyMerr: Yes i can. I show 20 cells. reach bottom, load another 20 cells from backend and than reach bottom again

freak4pc: But then the offset would change since you added more cells, so it would be like "scrolling" and would change back to false. At least that's what it seems to me, I didn't try it.

freak4pc: I thought the offset would change but it actually wouldn't change by just adding cells, I imagine. I don't have the time to try this - @jdisho did you test this by any chance?

jdisho: Sure I did @freak4pc and works as expected.

freak4pc: Ok, @MortyMerr - did you actually test your hypothesis? I wrote something like this as well and remember adding `distinctUntilChanged` or it was too sporadic.

MortyMerr: @freak4pc I just didn't get your point from the first time. Yes, it will work as you described. I added `disntinctUntilChanged`

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,52 @@
...
```
freak4pc: This regressed back (the scrollview thing)

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,52 @@
...
+final class ReachedBottomViewController: UITableViewController {
+    private let dataSource = Array(stride(from: 0, to: 28, by: 1))
+    private let identifier = "identifier"
+    private let disposeBag = DisposeBag()
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: identifier)
+        tableView.rx.reachedBottom(offset: 40)
+            .bind { print("Reached bottom") }
```
freak4pc: If you're not performing actual binding to something, I'd use `subscribe` as the intent is more explicit. 

```suggestion
            .subscribe { print("Reached bottom") }
```

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,31 @@
...
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the bottom of the UIScrollView is reached.
+     - parameter offset: A threshhold indicating the bottom of the UIScrollView.
+     - returns: ControlEvent that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
+        let source = contentOffset.map { contentOffset in
+            let visibleHeight = self.base.frame.height - self.base.contentInset.top - self.base.contentInset.bottom
```
freak4pc: I know we've had this discussion already below but - I still think it's best to be safe and add a `[weak base]` or `[weak self]` here.

The fact you don't have a retain cycle is related to the current implementation of `contentOffset`. There's no guarantee a retain cycle won't be create if the underlying implementation changes in a later version of RxSwift. 

My opinion is that external code should protect itself and not count on the core code. WDYT?

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
MortyMerr: `bind` is official RxCocoa operator. The only difference with `subscribe(onNext:)` - it will crash the app with in case of an error in debug mode. I can't see any difference in code clearance here.  I would prefer to leave `bind`.

freak4pc: I asked using `subscribe`, not `subscribe(onNext:)`. That's what all other examples in this repo use, please change to it.

freak4pc: PS bind isn't an "operator". It's an `alias` to `bind`.

jdisho: But you are not binding to something here. `subscribe` suits better to the context.

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: > external code should protect itself and not count on the core code

I think it's against all principle of abstraction. If core changing -> you change your system, which is based on the core. This is how things work from my experience. We don't wrap swift core functions in our own, because "one day API will be changed, so we can't trust swift"

If you want, I can add a unit test on the reference cycle here. It will fail if `contentOffset` changes.

freak4pc: Sorry, but I fail to understand why you're so stubborn about this specific thing. There is no performance gain, just added safety.

When you wrap Swift, you're wrapping a black box. You have no idea to know what's going on inside to know if the inner code keeps you safe from a retain cycle or not - so when wrapping 3rd party code, you should treat it as that same black box. 

It's somewhat bad-practice to base an implementation detail related to memory safety only on the fact you're intimate with the specific implementation details of the wrapped implementation.

It doesn't make too much sense to me. If there was some performance hit here or something, I'd definitely at least understand where you're coming from, but here it just seems you don't want to add it for now real apparent reason.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
MortyMerr: Lol, guys, maybe I didn't understand documentation right. Let's try to read it together
```
       /**
        Subscribes an element handler to an observable sequence.
        In case error occurs in debug mode, `fatalError` will be raised.
        In case error occurs in release mode, `error` will be logged.
    
        - parameter onNext: Action to invoke for each element in the observable sequence.
        - returns: Subscription object used to unsubscribe from the observable sequence.
        */
        public func bind(onNext: @escaping (Self.Element) -> Void) -> Disposable
```
I can see this in RxCocoa sources. 
>  Action to invoke for each element in the observable sequence.

What's the problem?

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: So, you won't merge it, if i won't add `[weak self]`?

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
freak4pc: I prefer not spending more time bike-shedding this. This is how other examples in this repo do it, please change your examples. Thanks.

Source/RxCocoa/UIScrollView+reachedBottom.swift
freak4pc: This isn't your repo, it's a community repo, everyone can make modifications. 

Even if I'd merge this code, it doesn't mean I can't add the capture group later. 

Regardless, you seem to be "fighting" with myself and @jdisho on every single comment we write, after basically taking over this PR which was originally written by @jdisho. 

Contributing to OSS is awesome, and I'm a huge fan of people who do so such as yourself, but you should also be a good OSS citizen. In this specific case you're fighting over nothing.

I prefer the added safety and taking care of "our own", which is not useless, even if you think so. UIKit implementation also changed and leaked several times, we need to do what we can so the code is independent in that case.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
MortyMerr: @freak4pc sorry, but i don't think it's `bike-shedding`. I came here to contribute and to make someone's life easier. I fulfill guidelines and swiftlint, added tests, docs and examples.
And here you come with secret rules, which i can get only for you. Official docs tell me: "hey man, here is function, use it". You tell me: "Sorry, it's banned". Of course, I will do like you said, but, where can i find this rule "not use `bind(onNext:)`"? Maybe you should also add it to contributing.md? Such comments about secret subjective rules make contributing very complicated.
Thanks :)

freak4pc: 1. You're not using `bind(onNext:)`, you're using `bind {}`.
2. `bind` operators are an alias to `subscribe(onNext:)` (read the source code). The former is meant to specifically be used for binding to UI elements, which isn't what you're doing here. It's a contextual addition, not a practical one. 
3. As I said for 2 times already - if all examples in this repo use one variation, we want to keep it aligned throughout all examples, regardless of your own personal preference. I'm not going to add every tiny thing to the contribution guides, that's why PR reviews exist. 

If you're not willing to make the modifications, we can do it.

MortyMerr:
@freak4pc  @jdisho Didn't think that not agreeing with your every word and providing arguments for my position is - "fighting". It seems like you are blaming me in some kind of toxicity. To prevent it - I will commit all your comments. No more discussions, thank you and sorry

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
MortyMerr: bind(onNext:) == bind {} Trailing closure syntax.

MortyMerr: Change it

freak4pc:
I'm not blaming over anything, I just thing you're fighting over the wrong things :) We both asked for minor details (first is added safety - matter of taste and personal experience, 2nd is usage of bind-vs-subscribe which is the current usage inside this repo), but you're very stubborn with using your own preference and not this repo's preferences and existing code uses.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
freak4pc: Sorry, I thought that it's bind and subscribe. Basically since there are no errors inside UI code anyways, it doesn't make a difference. Also, as I mentioned before, that's how other code in this repo does it. Thanks.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
MortyMerr: Thanks, fixed

Source/RxCocoa/UIScrollView+reachedBottom.swift
MortyMerr: Fixed, thanks

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,52 @@
...
+import RxSwift
+import RxCocoa
+import RxSwiftExt
+import PlaygroundSupport
+import UIKit
+
+/*:
+ ## reachedBottom
+
+ The `reachedBottom` operator provides a sequence that emits every time `UIScrollView` is scrolled to the bottom.
```
freak4pc:
```suggestion
 The `reachedBottom` operator provides a sequence that emits every time the `UIScrollView` is scrolled to the bottom.
```

Source/RxCocoa/UIScrollView+reachedBottom.swift
```diff
@@ -0,0 +1,31 @@
...
+import RxCocoa
+
+public extension Reactive where Base: UIScrollView {
+    /**
+     Shows if the bottom of the UIScrollView is reached.
+     - parameter offset: A threshhold indicating the bottom of the UIScrollView.
+     - returns: ControlEvent that emits when the bottom of the base UIScrollView is reached.
+     */
+    func reachedBottom(offset: CGFloat = 0.0) -> ControlEvent<Void> {
+        let source = contentOffset.map { [base] contentOffset in
```
freak4pc: I'm pretty sure capturing `base` this way is basically like not doing anything :) 
Either remove it completely like you wanted or make it a `weak base`.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
freak4pc: Minor text fix here.

freak4pc:
I think we also missed updating README.md.

Playground/RxSwiftExtPlayground.playground/Pages/UIScrollView.reachedBottom.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,52 @@
...
+import RxSwift
+import RxCocoa
+import RxSwiftExt
+import PlaygroundSupport
+import UIKit
+
+/*:
+ ## reachedBottom
+
+ The `reachedBottom` operator provides a sequence that emits every time the `UIScrollView` is scrolled to the bottom.
```
freak4pc: Also just realised reachedBottom isn't an operator...

MortyMerr: Lol, yes. Can i open additional Pr on that & readme.md?

freak4pc: Just pushed these on master. Let me know if you have any notes for it.

MortyMerr: Looks perfect, thank your for comments

freak4pc:
@jdisho I only now noticed the extension is directly on `UIScrollView` so you're correct, there can't even theoretically be a retain cycle. Also the `[base]` you added there is the same as not having anything (unless you're doing [weak base] but that doesn't seem to be needed here). 

I'm an opinionated person but I think it's important to be fair and also say when you're wrong ;-) I'll make the adjustment.

MortyMerr:
@freak4pc yes it does. Hope, next time we can do better :)
As for `[base]` as i know capture list create copy of reference. So it doesn't capture self, but capture copy of `base` otherwise. Am i wrong?

freak4pc:
Yup, just that `self` is `Reactive` so it doesn't matter in this case.

MortyMerr:
But if `self` is class, [base] will help, won't it?

MortyMerr:
@freak4pc there is another question for you (i just don't know, where to ask)
Why only `ObservableType` has `func materialize() -> Observable<Event<E>>`? Why `Single` or `Completable` doesn't?

PR #175, opened 2018-08-21
Update README.md

ericmarkmartin:
Add instructions to install `RxSwiftExt/RxCocoa` subspec

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -35,6 +35,18 @@ If you use Swift 2.x:
 pod "RxSwiftExt", '1.2'
 ```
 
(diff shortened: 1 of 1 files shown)


freak4pc:
Thanks for this Eric! Saw your DMs earlier. 
I actually think it would be better to have the default behavior (e.g. `pod 'RxSwiftExt'`) do both RxSwift and RxCocoa extensions, and have people who want _only_ RxSwift extensions simply use the subpsec. 

I'll make a PR around this in the next few days, so if you don't mind I think it's better to close this one for now :) Thoughts?

ericmarkmartin:
Good by me. Thanks for the quick feedback as always

PR #181, opened 2018-09-22
maybe: add a ifEmpty(switchToSingleWithError:) extension

twittemb:
This commit adds an extension the trait "Maybe" that transforms it into a Single
whenever it is empty (ie completes without succeed).

The difference with the already existing "ifEmpty" function is that the Single
will immediately emmit the specified Error.

It allows to consider the case when the Maybe is empty as an applicative error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ master
 - added `mapMany` operator
 - added `toSortedArray` operator
 - rolled `map(to:)` back to `mapTo(_:)` for `SharedSequenceConvertibleType`
+- added `ifEmpty(switchToSingleWithError:)` to the trait `Maybe`
 
 3.3.0
 -----
--- a/Playground/RxSwiftExtPlayground.playground/Pages/ifEmptySwitchToError.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/ifEmptySwitchToError.xcplaygroundpage/Contents.swift
@@ -1,0 +1,39 @@
+/*:
+ > # IMPORTANT: To use `RxSwiftExtPlayground.playground`, please:
+
+ 1. Make sure you have [Carthage](https://github.com/Carthage/Carthage) installed
+ 1. Fetch Carthage dependencies from shell: `carthage bootstrap --platform ios`
+ 1. Build scheme `RxSwiftExtPlayground` scheme for a simulator target
+ 1. Choose `View > Show Debug Area`
+ */
+
+//: [Previous](@previous)
+
+import RxSwift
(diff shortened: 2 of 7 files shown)


twittemb:
Hi here.

I don't know if that kind of operator could be in the RxSwift core project. In my humble opinion it is too specific for that, but I live the decision to you guys.

See ya.

twittemb:
Hi @freak4pc 

Thanks for your feedback. 

Transforming a Maybe into a Single if the Maybe is empty is already part of the RxSwit core. 

It allows to consider a Maybe's emptiness as a edge case of the normal workflow. It can be useful if you use a library that exposes a Maybe and Its emptiness really means something in your specific case ... And by something I mean a meaningful value for the rest of your workflow (emptiness is not an option for you).

It is the same idea with my proposal, but the emptiness would represent an Error (this is the meaningful value) in the user/developer's workflow.

In my humble opinion it is a complement of what already exists in the RxSwift core 😀.

I leave the final word to you of course.

Thanks again.

twittemb:
Hi guys ... any new thoughts on this topic ? @freak4pc

freak4pc:
Thinking of this more I still find the edge case too narrow/niche.

Treating an empty Maybe as a SingleEvent.error feels way too specific in my humble opinion. (It also feels odd in my opinion to cross the two types this way). 

Unless there is a more broad acceptance / demand for this kind of operator, I would personally be -1 to adding it.

Inviting @RxSwiftCommunity/contributors for additional opinions.

twittemb:
ok ... no prob with that. May I close this PR right away ?

freak4pc:
Maybe let’s wait to see if other contributors might find this valuable first. If there’s no responses for a few days, we can close this potentially.

jegnux:
@twittemb I agree with @freak4pc. It's not something too widely used imho to deserve it's own operator. Especially because it's not really complicated to write this when needed.

By the way, you can also easily transform the emptiness into an error, simply by converting it to an `Observable` and then again to a `Single`, because by default, Rx emit an `RxError.noElements` error if a `Single` sequence is empty.  :
```
Maybe<Int>
    .create { observe -> Disposable in
        observe(.completed)
        return Disposables.create()
    }
    .asObservable()
    .asSingle()
    .subscribe { print($0) }
```

If we would want to add a shortcut here, I would suggest to implement `asSingle()` directly on `PrimitiveSequence` in order to avoid the intermediate `asObservable()` but that's all. 

Then, swapping `RxError.noElements` with an other error would remains the responsibility of an other operator (typically `catchError`)

twittemb:
Hi.

Thanks for the feedback ... And the explanations. I close the PR.

Have a good day.

PR #185, opened 2018-10-12
Handle thrown errors in filterMap and flatMapSync

sgtsquiggs:
RxSwift's map, filter, and flatMap all catch (and transmit) errors via `throw`, so I have added this functionality to flatMapSync and filterMap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 Changelog
 =========
 
-master
+3.4.0
 -----
 - Fix `withUnretained` so it allows proper destructuring
 - added `mapMany` operator
 - added `toSortedArray` operator
 - rolled `map(to:)` back to `mapTo(_:)` for `SharedSequenceConvertibleType`
+- added `filterMap` and `flatMapSync` overloads for throwable transforms
 
 3.3.0
 -----
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "ReactiveX/RxSwift" "4.2.0"
+github "ReactiveX/RxSwift" "4.3.1"
--- a/RxSwiftExt.podspec
+++ b/RxSwiftExt.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "RxSwiftExt"
-  s.version      = "3.3.0"
+  s.version      = "3.4.0"
   s.summary      = "RxSwift operators not found in the core distribtion"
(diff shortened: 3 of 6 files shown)


freak4pc:
Thanks for the contrib Matthew :) 
Do you have any practical use case where you needed this? I don't feel super comfortable breaking public interface that way (specifically for filterMap which is being used quite heavily by people). If we do this it will have to be with some deprecation fix-it. Not sure if its worth it. Thoughts?

sgtsquiggs:
Wouldn't this be backwards compatible? Every existing transform function w/o `throw` would still work.

I have use cases for `map` throwing errors (map-with-error isn't that uncommon is it?). If I want to `filter` in addition (and easily w/ `filterMap`) I have to instead `filter` first then `map` the remaining. This is especially annoying if I'm using a switch statement since the states I `filter`'d to error now exist in my map but those cases will never fire.

A simplified example:

```swift
struct SomeThirdPartyConnectionStatusEnum {
    case connecting
    case connected
    case disconnecting
    case disconnected
    case errored(Error)
}
```

```swift
connectionStatusObservable
    .map { status -> Bool? in
        switch status {
        case .connected:
            return true
        case .disconnected:
            return false
        case let .error(error):
            throw error
        default:
            return nil
    }
    .filterNil()
```
vs
```swift
connectionStatusObservable
    .filterMap { status -> FilterMap<Bool> in
        switch status {
        case .connected:
            return .map(true)
        case .disconnected:
            return .map(false)
        case let .error(error):
            throw error
        default:
            return .ignore
    }
```

freak4pc:
I would be OK adding this with a proper deprecation warning, and probably a version bump as well.

sgtsquiggs:
Since the new method will be an overload of the existing method there will be a "fun" issue with deprecation. All non-throwing closure blocks will by default use the deprecated non-throwing method, resulting in false-positives. 😑

MortyMerr:
@freak4pc what do you think about it? :)

freak4pc:
Yeah, it looks good. I'm a bit worried about overloads not working well, so we might want to cover this with tests and of course rebase on top of the latest version.

MortyMerr:
I will communicate with @sgtsquiggs and if he doesn't mind will work on that by myself

sgtsquiggs:
Feel free to continue the good work 😎

MortyMerr:
Thanks :)

MortyMerr:
Closing cause changes were already merged

PR #188, opened 2018-10-23
[feature/applySingle] implement apply for SingleTrait

MortMerr:
Don't know, should i write separate test for that, cause implementation are equal

### DIFF
--- a/Source/RxSwift/apply.swift
+++ b/Source/RxSwift/apply.swift
@@ -15,3 +15,10 @@ extension ObservableType {
         return transform(self.asObservable())
     }
 }
+
+extension PrimitiveSequenceType where TraitType == SingleTrait {
+    /// Apply a transformation function to the Single.
+    func apply<T>(_ transform: (Single<Self.ElementType>) -> Single<T>) -> Single<T> {
+        return transform(self.primitiveSequence)
+    }
+}


freak4pc:
Yeah - you should add tests. Mainly because the tests will describe Single, so if that specific implementation changes one day, the tests will still cover it.

MortMerr:
@freak4pc added test. Please, rereview

freak4pc:
Thanks @MortMerr !

Two more things:

1. Please add a CHANGELOG entry.
2. Might be nice adding an example to the playground page that uses Single.

MortMerr:
@freak4pc added

Playground/RxSwiftExtPlayground.playground/Pages/apply.xcplaygroundpage/Contents.swift
```diff
@@ -44,4 +44,36 @@ example("apply a transformation") {
...
 }
 
+/*:
+ ## apply
+
+ The `apply` operator takes a transformation function `(Single) -> Single` and applies it to the stream. The purpose of this operator is to provide syntactic sugar for applying multiple operators to the stream, while preserving the chaining operator structure of Rx.
+
+ */
+
+func addOne(input: Single<Int>) -> Single<String> {
```
freak4pc: Hey, sorry for nit-picking but this isn't exactly what I mean. Just adding a few example using `Single` in the existing sample on top would be enough. No need for an additional title and more code blocks :)

MortMerr:
@freak4pc is it ok now?

PR #189, opened 2018-11-03
[Proposal] flatMapMany Operator

touyu:
### Motivation for inclusion

For example, if you want to perform asynchronous processing on each element of observable array, write as follows:

```swift
// `fetch()` is asynchronous processing and returns observable value.
theObservable
  .flatMap { array in 
    let new = array.map { $0.fetch() }
    return Observable.from(new).merge().toArray()
  }
```
I want to write the above code more concisely:
```swift
theObservable
  .flatMapMany { $0.fetch() }
```

### Example of use
It is described above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ master
 - rolled `map(to:)` back to `mapTo(_:)` for `SharedSequenceConvertibleType`
 - added `unwrap()` operator for SharedSequence
 - added `apply(_:)` for `Single`
+- added `flatMapMany` operator
 
 3.3.0
 -----
--- a/RxSwiftExt.xcodeproj/project.pbxproj
+++ b/RxSwiftExt.xcodeproj/project.pbxproj
@@ -175,6 +175,8 @@
 		D1EB19E4218DBED4008FC64A /* FlatMapManyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1EB19E2218DBE68008FC64A /* FlatMapManyTests.swift */; };
 		D1EB19E5218DBED5008FC64A /* FlatMapManyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1EB19E2218DBE68008FC64A /* FlatMapManyTests.swift */; };
 		D1EB19E6218DBED5008FC64A /* FlatMapManyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1EB19E2218DBE68008FC64A /* FlatMapManyTests.swift */; };
+		D1EB19E7218DCB13008FC64A /* flatMapMany.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1EB19E0218DB892008FC64A /* flatMapMany.swift */; };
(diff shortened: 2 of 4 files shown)


freak4pc:
Hey @touyu - Thanks for the suggestion and implementation.
I'm personally quite torn and currently adding against such an operator. 

The issue is - as opposed to `mapMany` on `map` - `flatMap` creates additional resources / on-air streams. Using an operator that can create so many resources in a way that's "masked" into such a simplistic operator feels very very dangerous in the hands of 90% of users. 

Unless there is demand from community, or other opinions, I'd currently be against merging this unfortunately. 

Inviting @RxSwiftCommunity/contributors to comment their thoughts. 

Thanks again!

jdisho:
Nice idea! Looks good to me!

freak4pc:
@jdisho Please re-read my message.

`mapMany` is 100% fine. It does a simple non-escaping mapping from `[T]` to `[R]`. 

`flatMap`, on the other hand, creates new subscriptions (e.g. resources) on every run. To "hide" this fact under a simplistic operator such as `flatMapMany` is dangerous since people will not be aware of what this does. 

```swift
array.
 flatMapMany { doLongProcess(with: $0) } 
```

Now thing of this:
```
array is [1, 2, 3] -> 3 running calls
array changes to [4, 5, 6] -> 6 running calls
array emits arbitrary 20 calls -> 26 running calls
```

This kind of exponential increase in resources will be hidden inside this operator where the consumer of it (a developer) wouldn't usually know or understand the performance cost of it (not would be able to debug it without going into the operator implementation). 

And on, and on, there is no disposition (since its not a flatMapLatest), unless the inner requests completes. Also, his implementation includes a `toArray` which means its too opinionated (it doesn't just flatMaps, it also must be a finite sequence). 

In my eyes it's too niche, too dangerous for usage and doesn't provide too many benefits over writing the full expression where you have control and explicitness.

jdisho:
Thank you for the explanation 😊 @freak4pc

jdisho:
What about this @freak4pc ?

```swift 
public func flatMapMany<O>(_ selector: @escaping (Self.E.Element) -> O) -> Observable<[O]> {
        return flatMapLatest { collection -> Observable<[O]> in
            let new = collection.map(selector)
            return Observable.just(new)
        }
    }
```

freak4pc:
I’m not sure how it’s solving the issue or is different than mapMany. My opinion is that in general this kind of helper would be much more harm than good.

touyu:
Thank you for confirming this PR.

Is it still dangerous with the following code?

```swift
public func flatMapMany<O>(_ selector: @escaping (E.Element) throws -> O) -> Observable<[O.E]> where O: ObservableType {
    return flatMap { collection -> Observable<[O.E]> in
        let new = try collection.map(selector)
        return Observable.zip(new)
    }
}
```

freak4pc:
Hey all - I'm currently leaning towards closing this PR. Even though I'm very happy and appreciative of the work put into this, it's just too risky to mask flatMap(s) and resource-creation that way IMO. There also doesn't seem to be another other requests for this sort of operator in this history of Ext. If there's no other feedback in the next 24hrs or so, I'll be closing this. Thanks !

PR #191, opened 2018-11-12
Implement bufferWithTrigger.

patmalt:


freak4pc:
Hey @patmalt - Thanks for your contribution :) 

Just a few notes: 

1. When adding new operators, it's better to first come on Slack or Open an Issue to discuss the operator, how it's designed, and how it would be used, before doing any actual coding. 

2. Since you already wrote code and made a PR (great job!), we can't really consider the PR without at least some description - you left your PR entirely empty. Please write a description of what this does, why it's needed, and usage examples inside the body of your PR.

Thank yoU!

patmalt:
We are using this operator for our project, and we needed to write tests for it as well. Since we were already doing the work for this operator, I figured I would submit it and everyone could benefit. Sorry for skipping the discussion part of contributing. We created this operator after looking at `pausableBuffered` but realized it did not meet our needs.

The operator `accumulatingBuffer` collects the elements emitted by the source observable, and emits them as an array when the boundary observable emits. `accumulatingBuffer` can be used to gather items together, and then process them at some other time.

freak4pc:
Thanks! 
So what you're saying is the stream will buffer elements and spit them out whenever there's an emission from a trigger, basically? I don't know pausableBuffered so much, but isn't that exactly what it currently does ?

freak4pc:
Thanks! 
So what you're saying is the stream will buffer elements and spit them out whenever there's an emission from a trigger, basically? I don't know pausableBuffered so much, but isn't that exactly what it currently does ?

patmalt:
No, `pausableBuffered` pauses the elements of the source observable sequence based on the latest element from the second observable sequence. Our use case does not care about pausing of observables, we want to collect everything. Additionally, the pauser's type must be Bool, which limits which observables that can do the signaling.

freak4pc:
I don't know much about the usage / etc so I'll need to carve out some time for this, will have to wait for the weekend for me but if anyone else wants to experiment with and review this please do: @RxSwiftCommunity/contributors !

Tests/RxSwift/AccumulatingBufferTests.swift
```diff
@@ -0,0 +1,92 @@
...
+        let res = scheduler.start(disposed: 1000) {
+            underlying.accumulatingBuffer(boundary.asObservable())
+        }
+
+        XCTAssertEqual(res.events, [
+            next(201, [2]),
+            next(301, [3, 4]),
+            next(401, [5, 6, 7]),
+            next(500, [8]),
+            completed(500)
```
freak4pc: My question here is - is a `completed` really considered a boundary emission? that kinda seems like an unwanted side-effect. Basically a developer could do `take(1)` so I guess it's fine, but was that the desired behavior?

freak4pc:
IRT the name - What do you think about something like `triggeredBuffer` or `bufferWithTrigger` (there is `bufferWithTime` in RxJava, for example). I kinda like the verbosity of it.

Tests/RxSwift/AccumulatingBufferTests.swift
patmalt: Yes, this gives the observer of this new operator one last chance to receive buffered events.

CHANGELOG.md
```diff
@@ -14,7 +14,11 @@ master
 - rolled `map(to:)` back to `mapTo(_:)` for `SharedSequenceConvertibleType`
 - added `unwrap()` operator for SharedSequence
 - added `apply(_:)` for `Single`
+<<<<<<< HEAD
```
freak4pc: you left merge markers

freak4pc:
You left merge markers by accident, I believe

CHANGELOG.md
patmalt: I see that, fixing now!

Readme.md
```diff
@@ -80,8 +80,12 @@ These operators are much like the RxSwift & RxCocoa core operators, but provide
 * [Observable.fromAsync](#fromasync)
 * [Observable.zip(with:)](#zipwith)
 * [withUnretained](#withunretained)
+<<<<<<< HEAD
 * [count](#count)
 * [partition](#partition)
+=======
```
freak4pc: More over here

PR #192, opened 2018-11-22
add swift package manager support.

imkerberos:
#trivial
Add "Package.swift" to supporting Swift Package Manager.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,19 @@ jobs:
       - run:
           name: Push Podspec to Trunk
           command: pod trunk push --skip-tests --allow-warnings
+  "RxSwiftExt SPM"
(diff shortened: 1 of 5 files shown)


vzsg:
An example config could look as simple as this:

```yaml
version: 2
jobs:
  "RxSwiftExt SPM":
    macos:
      xcode: '9.4.0'
    steps:
      - checkout
      - run: swift build
      - run: swift test
```

vzsg:
On a second thought, it might be better to extend the RxSwiftExt Test job with one more named run section, instead of introducing a separate job. It took me too long to realize how the script works.

Try adding this after the "Test tvOS" step:

```yaml
      - run:
          name: Test Swift Package Manager
          command: set -o pipefail && swift build && swift test
```

And sorry for sidetracking you.

freak4pc:
Resolved in #230

PR #193, opened 2018-11-27
Partition operator

freak4pc:
This is a new operator we've been using in our codebase that seems generic and useful, and I think would be a good fit here :) 

Partition would split a stream into two streams - one with elements matching the predicate, and one with elements not with matching the predicate.

The name partition came from Haskell: http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List.html#v:partition

This lets you do things like 

```swift
let numbers = Observable.of(1, 2, 3, 4, 5, 6)

let (evens, odds) = numbers.partition { $0 % 2 == 0 }

_ = evens.debug("even").subscribe() // emits 2, 4, 6
_ = odds.debug("odds").subscribe() // emits 1, 3, 5

// Another example:

let currentState: Observable<State> = ...

let (isLoading, isIdle) = currentState.partition { $0 == .loading }
```

Would appreciate your opinion :)

### DIFF
--- a/Source/RxSwift/partition.swift
+++ b/Source/RxSwift/partition.swift
@@ -1,0 +1,27 @@
+//
+//  partition.swift
+//  RxSwiftExt
+//
+//  Created by Shai Mishali on 24/11/2018.
+//  Copyright © 2018 RxSwiftCommunity. All rights reserved.
+//
+
+import RxSwift
+
+public extension ObservableType {
+    /**
+     Partition a stream into two separate streams of elements that match, and don't match, the provided predicate.
+
+     - parameter predicate: A predicate used to filter matching and non-matching elements.
+
+     - returns: A tuple of two streams of elements that match, and don't match, the provided predicate.
+    */
+    func partition(_ predicate: @escaping (E) throws -> Bool) -> (matches: Observable<E>, nonMatches: Observable<E>) {
+        let stream = self.share()
+        let hits = stream.filter(predicate)
+        let misses = stream.filter {
+            return !(try predicate($0))
+        }
+        return (hits, misses)
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Changelog
 master
 -----
 
-Nothing yet!
+- added `partition(_:)` operator
 
 3.4.0
 -----
--- a/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
+++ b/Playground/RxSwiftExtPlayground.playground/Pages/Index.xcplaygroundpage/Contents.swift
@@ -44,6 +44,7 @@
  - [ofType()](ofType) operator, filters the elements of an observable sequence, if that is an instance of the supplied type.
  - [toSortedArray()](mapMany) operator, converts an Observable into another Observable that emits the whole sequence as a single array sorted using the provided closure and then terminates.
  - [count(predicate)](count) operator, counts the number of items emitted by an Observable. If predicate exists, then counts the number of items satisfying it.
+ - [partition](partition) operator, partition a stream into two separate streams of elements that match, and don't match, the provided predicate.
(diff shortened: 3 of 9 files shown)


freak4pc:
If anyone would like to do a quick review just for sanity check, that'd be great. Wouldn't want to merge it without another set of eyes on the code itself. @RxSwiftCommunity/contributors

mkko:
> A little bikeshedding: there already is a `partition(by:)` method in the Swift Standard Library, with a very different meaning.
> 
> What about `split`? It even makes the documentation read more easily:
> 

Fair point, but I think `split` also has a different meaning. I would read it as splitting it into many pieces in such a way as strings do. Also, I think the functionality of `partition(by:)` is actually closer to this one.

> BTW My original idea was split (or branch), but I felt it's a bit simplisitc and partition made more sense.


As for `branch`, I think it sounds more exclusive. I think the name `partition` describes the functionality well.

groue:
I think that it had to be said. Now you and @mkko can make a better decision 👍

freak4pc:
Merging this for now. We can always rename and deprecate if there is strong community indifference later on. Thanks all!

malaba:
Question: in the current implementation the `predicate` is evaluated twice for each next event. Would it be possible to evaluate the predicate only once ? And then emit on either `hits` or `misses` ?

freak4pc:
@malaba - Yeah, it would filter each element on both "partitions". 
I don't think there's a good way to do this in a single iteration. 

The only way I can think of is with something like a relay but that would be much worse and will break a lot of the composability we're gaining here. 

Basically it's not a real loss, because someone who wants to do this kind of pattern will usually do two filters, anyways, so this is basically syntactic sugar over the same strategy.

Source/RxSwift/partition.swift
```diff
@@ -0,0 +1,27 @@
...
+public extension ObservableType {
+    /**
+     Partition a stream into two separate streams of elements that match, and don't match, the provided predicate.
+
+     - parameter predicate: A predicate used to filter matching and non-matching elements.
+
+     - returns: A tuple of two streams of elements that match, and don't match, the provided predicate.
+    */
+    func partition(_ predicate: @escaping (E) throws -> Bool) -> (matches: Observable<E>, nonMatches: Observable<E>) {
+        let stream = self.share()
```
jegnux: Just wondering, because I often do the same in my custom operators, is it really fine to `share()` an observable under the hood? It might be an unexpected side effect don't you think?

freak4pc: I think it's the common path - no "harm" done from doing that for the common use cases. I can't really imagine a use case where you wouldn't want to share though. 


Usually I do agree that it's a bit weird to use it under the hood but for something like this, I feel it makes sense. If there are any issues with this we can expand it later on, with some flag etc.

mkko:
> Question: in the current implementation the `predicate` is evaluated twice for each next event. Would it be possible to evaluate the predicate only once ? And then emit on either `hits` or `misses` ?

Actually a really good point. I wonder if the following would work. I think we could evaluate the predicate and put it in to the observable.

```swift
func partition(_ predicate: @escaping (E) throws -> Bool) -> (matches: Observable<E>, nonMatches: Observable<E>) {
    let stream = self.map { ($0, try predicate($0)) }.share()

    let hits = stream.filter { $0.1 }.map { $0.0 }
    let misses = stream.filter { !$0.1 }.map { $0.0 }
    return (hits, misses)
}
```

mkko:
...and I'm just a little worried that people write something into the filter with a side-effect, e.g. put the item to a collection of known/existing things and check the collection on the next pass.

Not a good argument for though.

freak4pc:
Uhm I don't think it's costly :) We can definitely make that adjustment but it just feels like premature optimization. These terms should be very simple.

Anyways, I'm all good for adding that if you'd like but not sure if it provides any significant performance benefit for 99% of cases, up to you. I really don't mind :)

freak4pc:
Added the discussed change in #195.

mkko:
> Question: in the current implementation the `predicate` is evaluated twice for each next event. Would it be possible to evaluate the predicate only once ? And then emit on either `hits` or `misses` ?

This was resolved in #195. Thank you for pointing this out!

PR #195, opened 2018-11-28
Partition - do a single pass for predicate

freak4pc:
Optimization discussed in #193.


mkko:
That was quick. It took me longer to write and not even finish that I don't have the time until later this week :-D

PR #199, opened 2019-03-06
[feature/fromAsync] add from async for single

MortyMerr:
I think the current implementation of `fromAsync` might be a little more useful. Completion handlers used to be `(result: T, error: CustomError)` closures. Also completion handler should emit only once. So i used `Single` to implement that

### DIFF
--- a/Source/RxSwift/fromAsync.swift
+++ b/Source/RxSwift/fromAsync.swift
@@ -66,3 +66,70 @@ extension Observable {
         return { (a: A, b: B, c: C, d: D, e: E, f: F, g: G, h: H, i: I, j: J) in Observable.fromAsync(curry(asyncRequest)(a)(b)(c)(d)(e)(f)(g)(h)(i)(j)) }
     }
 }
+
+public enum FromAsyncError: Error {
+    /// Both result & error can't be nil
+    case inconsistentCompletionResult
+}
+
+public extension PrimitiveSequenceType where TraitType == SingleTrait {
+    /**
+     Transforms an async function that returns data or error through a completionHandler in a function that returns data through a Single
+     - The returned function will thake the same arguments than asyncRequest, minus the last one
+     */
+    static func fromAsync<Er: Error>(_ asyncRequest: @escaping (@escaping (ElementType?, Er?) -> Void) -> Void) -> Single<ElementType> {
+        return .create { single in
+            asyncRequest { result, error in
+                switch (result, error) {
+                case let (.some(result), nil):
+                    single(.success(result))
+                case let (nil, .some(error)):
+                    single(.error(error))
+                default:
+                    single(.error(FromAsyncError.inconsistentCompletionResult))
+                }
+            }
+            return Disposables.create()
+        }
+    }
+
+    static func fromAsync<A, Er: Error>(_ asyncRequest: @escaping (A, @escaping (ElementType?, Er?) -> Void) -> Void) -> (A) -> Single<ElementType> {
+        return { (a: A) in Single.fromAsync(curry(asyncRequest)(a)) }
+    }
+
+    static func fromAsync<A, B, Er: Error>(_ asyncRequest: @escaping (A, B, @escaping (ElementType?, Er?) -> Void) -> Void) -> (A, B) -> Single<ElementType> {
+        return { (a: A, b: B) in Single.fromAsync(curry(asyncRequest)(a)(b)) }
+    }
+
+    static func fromAsync<A, B, C, Er: Error>(_ asyncRequest: @escaping (A, B, C, @escaping (ElementType?, Er?) -> Void) -> Void) -> (A, B, C) -> Single<ElementType> {
+        return { (a: A, b: B, c: C) in Single.fromAsync(curry(asyncRequest)(a)(b)(c)) }
+    }
+
+    static func fromAsync<A, B, C, D, Er: Error>(_ asyncRequest: @escaping (A, B, C, D, @escaping (ElementType?, Er?) -> Void) -> Void) -> (A, B, C, D) -> Single<ElementType> {
+        return { (a: A, b: B, c: C, d: D) in Single.fromAsync(curry(asyncRequest)(a)(b)(c)(d)) }
+    }
+
+    static func fromAsync<A, B, C, D, E, Er: Error>(_ asyncRequest: @escaping (A, B, C, D, E, @escaping (ElementType?, Er?) -> Void) -> Void) -> (A, B, C, D, E) -> Single<ElementType> {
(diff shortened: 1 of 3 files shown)


patmalt:
+1

Tests/RxSwift/fromAsyncTests.swift
```diff
@@ -36,4 +44,47 @@ class FromAsyncTests: XCTestCase {
 
         XCTAssertEqual(observer.events, correct)
     }
+    
+    func testSingleResultEqualitySuccessCase() {
+        // given
+        let result = "result"
+        let expectedEvents = [next(0, result), completed(0)]
```
freak4pc: Please don't use the global deprecated `next` and `completed` functions

Cartfile.resolved
```diff
@@ -1 +1 @@
-github "ReactiveX/RxSwift" "4.4.0"
+github "ReactiveX/RxSwift" "4.4.1"
```
freak4pc: No reason to bump version here

Source/RxSwift/fromAsync.swift
```diff
@@ -66,3 +66,71 @@ extension Observable {
         return { (a: A, b: B, c: C, d: D, e: E, f: F, g: G, h: H, i: I, j: J) in Observable.fromAsync(curry(asyncRequest)(a)(b)(c)(d)(e)(f)(g)(h)(i)(j)) }
     }
 }
+
+public enum FromAsyncError: Error {
+    /// Both result & error can't be nil
+    case inconsistentCompletionResult
```
freak4pc: Is this relevant to Observable fromAsync as well ? Why can't result be optional and nil?

MortyMerr: Hi @freak4pc . Thank u for ur comments. Do u have any idea how should I handle a situation in which completionHandler have both error and success? I think throwing an inconsistent state error is the best option.

MortyMerr: If user want to make result optional type, fromAsync generic type would be double optional
`A??`, so it will work ok

Cartfile.resolved
MortyMerr: Sure, fixed

Tests/RxSwift/fromAsyncTests.swift
MortyMerr: I can't see any warning about deprecated methods usage. I also can't find deprecation in docs. Could u please attach some proof, that this functions were deprecated?

Source/RxSwift/fromAsync.swift
MortyMerr: I mean, outer optional goes for result optionality and inner optional is user defined type

Tests/RxSwift/fromAsyncTests.swift
mkko: It's a bit implicit, but here: [RxTest/Deprecated.swift](https://github.com/ReactiveX/RxSwift/blob/master/RxTest/Deprecated.swift#L35)

freak4pc: The deprecations are runtime deprecations but they've been there for a while.

Source/RxSwift/fromAsync.swift
mkko: I wouldn't treat empty result as an error nevertheless.

MortyMerr: Why? This is the way, how single works. It should fire exactly one event OR one error. I thunk throwing this documented error would be correct.
I agree about other comments, but here I am pretty sure in implementation

MortyMerr: Do u have any suggestions? Maybe better way to implement that?

freak4pc: @mkko I'm not too familiar with this operator, so if you've used it and have some opinions it would be highly helpful.

mkko: Sorry for the delay. I think that's a fair point, it does look like SingleAsync also expects a value and conformity should be prioritized.

I'm just thinking about a use case where you're not expecting a value, health/auth checks and such. Does it work if you define a `Single` with an optional value?

MortyMerr: I fully understand your concerns. But completion handler which has optional value should look like that:
```
typealias ResulType = String? 
func foo(completionHandler: (ResultType?, TestError?) -> Void)
// which is equal to
func foo(completionHandler: (String??, TestError?) -> Void)
```
We use outer optional to show that completion can be called with result OR with an error. And our inner optional goes for our actual result (which type is `String?`l).
About auth check - user should use `Completable` instead. But if he wants to use `Single` its `ElementType` type should be `Void`

mkko: Fair point. I think this looks good 👍

freak4pc: That makes sense - could we add a quick unit test for the optional case, just for sanity and coverage? I'll merge after that.

Tests/RxSwift/fromAsyncTests.swift
MortyMerr: Thanks, resolved

Tests/RxSwift/fromAsyncTests.swift
```diff
@@ -36,4 +44,47 @@ class FromAsyncTests: XCTestCase {
 
         XCTAssertEqual(observer.events, correct)
     }
+    
+    func testSingleResultEqualitySuccessCase() {
```
freak4pc: basically another test case where you do `let result: String? = "result"` would be great. The rest of the test should probably remain the same

Source/RxSwift/fromAsync.swift
MortyMerr: Just added additional test case with optional result type

MortyMerr:
@freak4pc is it ok now?

Tests/RxSwift/fromAsyncTests.swift
```diff
@@ -36,4 +44,70 @@ class FromAsyncTests: XCTestCase {
...
+            .asObservable()
+            .subscribe(observer)
+        scheduler.start()
+        // then
+        XCTAssertEqual(observer.events, expectedEvents)
+    }
+    
+    func testSingleOptionalResultEqualitySuccessCase() {
+        // given
+        let result: String? = "result"
```
mkko: I would test this against a `nil` value instead of string, since it would be the exceptional case. Otherwise this looks good to me.

mkko:
Just one thing about the testing of optional value. What do you think @freak4pc?

Tests/RxSwift/fromAsyncTests.swift
MortyMerr: Changed

MortyMerr:
Hey @mkko thank you for your review) Waiting for final approve

mkko:
Thank you, especially for the patience!

Issue #200, opened 2019-03-07
 'OSAtomicOr32OrigBarrier' was deprecated


xiao99xiao:
### Name and description
I'm using RxSwiftExt in a watchOS target app, and this warning has been existing for ages:
```
once.swift:30:23: 'OSAtomicOr32OrigBarrier' was deprecated in watchOS 3.0: Use atomic_fetch_or() from <stdatomic.h> instead
```

### Motivation for inclusion
Having this warning resolved by replacing it with the new `atomic_fetch_or()` or something would be great since Xcode no longer target below watchOS 3.0.

freak4pc:
Hey. This is related to the RxSwift version you use, and not RxSwiftExt itself.

Update to RxSwift 4.4.1 and you should be fine.

xiao99xiao:
@freak4pc I'm using RxSwift 4.4.1 and it still exists. The warning is  located at here:
https://github.com/RxSwiftCommunity/RxSwiftExt/blob/2204a685f7761244f311161f6c35ea0406488e17/Source/RxSwift/once.swift#L30

freak4pc:
Ah I see. Sorry about misunderstanding your question.

PR #209, opened 2019-05-13
`once`: Use NSRecursiveLock instead of deprecated OSAtomicOr32OrigBarrier

MortyMerr:
Resolve https://github.com/RxSwiftCommunity/RxSwiftExt/issues/200

@freak4pc mentioned that adding RxAtomic as dependency is an unwanted solution, so here is yet another `AtomicInt` implementation with few needed operators. Tests were just copied from `RxSwift` repo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Changelog
 =========
 
 - added `reachedBottom(offset:)` for `UIScrollView`
+- removed deprecated `OSAtomicOr32OrigBarrier`
 
 5.0.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ Changelog
 =========
 
 - added `reachedBottom(offset:)` for `UIScrollView`
-- removed deprecated `OSAtomicOr32OrigBarrier`
+- `once` now uses a `NSRecursiveLock` instead of the deprecated `OSAtomicOr32OrigBarrier`
 
 5.0.0
 -----
--- a/Source/RxSwift/once.swift
+++ b/Source/RxSwift/once.swift
@@ -25,12 +25,15 @@ extension Observable {
 	*/
 
 	public static func once(_ element: Element) -> Observable<Element> {
-		var delivered: UInt32 = 0
+        let lock = NSRecursiveLock()
+		var isDelivered = false
 		return create { observer in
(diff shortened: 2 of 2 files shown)


freak4pc:
Not sure I'd want to copy AtomicInt in its entirety just for this. Do you think `once` could be refactored with a regular `NSRecursiveLock` ?

MortyMerr:
@freak4pc do you mean something like this? Like:
```
let lock = NSRecursiveLock()
var delivered: UInt32 = 0
return create { observer in
    lock.lock()
    delivered = 1
    lock.unlock()
    if delivered == 0 {
        observer.onNext(element)
    }
    observer.onCompleted()
    return Disposables.create()
}
```

Source/RxSwift/once.swift
```diff
@@ -25,12 +25,15 @@ extension Observable {
...
 
 	public static func once(_ element: Element) -> Observable<Element> {
-		var delivered: UInt32 = 0
+        let lock = NSRecursiveLock()
+		var isDelivered = false
 		return create { observer in
-			let wasDelivered = OSAtomicOr32OrigBarrier(1, &delivered)
-			if wasDelivered == 0 {
+            lock.lock()
+			if !isDelivered {
```
freak4pc: I don't think you need this `if` statement anymore. Am I wrong about this?

MortyMerr: Maybe I missed the point, but who could I guarantee that event was only dispatched once without `if` statement?

MortyMerr: Hey @freak4pc could u please clarify? :)

freak4pc: Hey, sorry. This looks OK, can you just fix up the formatting? locking seems to be non-indented, etc

MortyMerr: Sorry, uses spaces instead of tabs...
Fixed

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ Changelog
 =========
 
 - added `reachedBottom(offset:)` for `UIScrollView`
+- removed deprecated `OSAtomicOr32OrigBarrier`
```
freak4pc: `once` now uses a `NSRecursiveLock` instead of the deprecated `OSAtomicOr32OrigBarrier`

PR #214, opened 2019-05-25
Simpify filterMap implementation, add throws to transform closure signature

MortyMerr:


TestErrors.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  TestErrors.swift
+//  RxSwiftExt-iOS
+//
```
freak4pc: This should be in the Tests folder (seems to br in the top level). 
I'm wondering if it's better to just have an error inside the test file like other operators.

Tests/RxSwift/filterMapTests.swift
```diff
@@ -50,18 +55,49 @@ class FilterMapTests: XCTestCase {
         subject.on(.next(1))
         subject.on(.next(2))
         subject.on(.next(3))
-        subject.on(.error(FilterMapTestError.error))
+        subject.on(.error(testError))
```
freak4pc: I kinda prefer the old way, where the operator tests had an error, so it's contained.

Tests/RxSwift/filterMapTests.swift
```diff
@@ -12,14 +12,20 @@ import RxSwift
...
 
-class FilterMapTests: XCTestCase {
-    func testIgnoreEvenAndEvenizeOdds() {
+    override func setUp() {
+        super.setUp()
+        scheduler = TestScheduler(initialClock: 0)
+    }
+
+    override func tearDown() {
+        scheduler = nil
```
freak4pc: not a huge deal but you don't really need this (a new one overrides the old one)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ Changelog
 =========
 
 - added `reachedBottom(offset:)` for `UIScrollView`
+- changed `filterMap(transform:)` callback on throwing
```
freak4pc:
```suggestion
- Simplify `filterMap(_:)` implementation and make callback throwing
```

freak4pc:
I'm not 100% sure we can make the closure throwing. It breaks backwards compatibility ... we might want to leave it as is until we bump a major version.

TestErrors.swift
MortyMerr: Move to Test folder, thanks.
> I'm wondering if it's better to just have an error inside the test file like other operators.
Honestly, i can see any point in boilerplate code with custom error for each class. TestErrors will reduce boilerplate. The approach is stolen from main repo

Tests/RxSwift/filterMapTests.swift
MortyMerr: Which pros does this approach have? I can see that contributor have to create his own dummy error in each class

Tests/RxSwift/filterMapTests.swift
MortyMerr: It's needed for resource deallocation after last test from this class
https://qualitycoding.org/xctestcase-teardown/

Tests/RxSwift/filterMapTests.swift
freak4pc: Contained tests, some operators may need errors that are more complex/specific, etc.

CHANGELOG.md
MortyMerr: Done, thanks

Tests/RxSwift/filterMapTests.swift
MortyMerr: But unless this error needed why not to use general dummy error?

freak4pc: It's just the way things currently happen inside every operator test in the repo. If you look at other tests in the repo you'll notice some have their own implementations and I'm not sure it's worth moving everything over. I don't super care to be honest.

TestErrors.swift
freak4pc: Don't have a huge preference, we can move it over. It's still not in the tests folder as far as I can tell though.

MortyMerr:
I trust your experience, so if u think we have to leave - let's leave.
But user could use `.filterMap` throwing version without changing his code. If you aware that someone will want downgrade RxSwiftExt version from throwing to non-throwing version - I can add throwing version as overload with later deprecation to the origina one(non-throwing). What do you think? :)

freak4pc:
I think it will conflict, can you try? (having a secondary overload with only a `throw` difference).

TestErrors.swift
MortyMerr: Pushed

Tests/RxSwift/filterMapTests.swift
MortyMerr: Me too. So i would prefer leave general dummy error for future test simplification, if you don't mind

freak4pc: Sure.

MortyMerr:
@freak4pc no conflict. Just pushed overload

freak4pc:
Thought about it some more - let's not do an overload, just add the `throws` property.

MortyMerr:
Sure, removed overload, change filterMap documentation

MortyMerr:
@freak4pc i added removing of `flatMapSync` here, cause it related to `filterMap` implementation

freak4pc:
I don't think it's necessarily related - can we remove flatMapSync in a separate PR? I mainly want the commits & PRs to be atomic so people can track down the changes backwards.

MortyMerr:
@freak4pc ok, reverted, will open separate PR

Tests/TestErrors.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  TestErrors.swift
+//  RxSwiftExt
+//
+//  Created by Anton Nazarov on 25/05/2019.
+//  Copyright © 2019 RxSwift Community. All rights reserved.
+//
+
+enum TestError: Error {
```
jdisho: Instead of using an one-case enum for errors, I would suggest using a `struct` instead.

MortyMerr: Enum can be extended with another cases if needed. This code was copied from main repo

jdisho: If more cases are needed than go with an `Enum`, but _now_ (especially in this PR) you need only one.

freak4pc: What would you gain from using a Struct here? I think it's worth just leaving as an enum because it's bound to be extended (as it spans the entire repo, not just one test)

freak4pc:
I'd want @jdisho to confirm he's OK with this as well, as long as he's reviewing here :)

PR #215, opened 2019-05-30
Deprecate Unwrap method in favour of compactMap

HugoSay:
RxSwift 5 added the CompactMap operator which overlaps with the current RxSwiftExt `unwrap()` method. 

Should we deprecate the use of the unwrap method in favour of the new compactMap naming?


Source/RxSwift/unwrap.swift
```diff
@@ -16,8 +16,17 @@ extension ObservableType {
...
-        return self.filter { $0 != nil }.map { $0! }
+        return self.compactMap({ $0 })
+    }
+
+    /**
+     Takes a sequence of optional elements and returns a sequence of non-optional elements, filtering out any nil values.
+
+     - returns: An observable sequence of non-optional elements
+     */
+    public func compactMap<T>() -> Observable<T> where Element == T? {
```
freak4pc: We probably shouldn't add this. Just deprecate the original. Thanks :)

HugoSay: Ok, I'll remove it then. 
Also I'll wait for the next release of RxSwift including compactMap on sharedSequence Single and Maybe In order to deprecate all the old code at once !
(https://github.com/ReactiveX/RxSwift/pull/1978)

Source/RxSwift/unwrap.swift
```diff
@@ -16,8 +16,8 @@ extension ObservableType {
 
      - returns: An observable sequence of non-optional elements
      */
-
-    public func unwrap<Result>() -> Observable<Result> where Element == Result? {
```
freak4pc, line 20: why did you change the signature? It's best to keep the same as it was, just add the deprecation, nothing else.

Issue #178, opened 2018-09-12
Extension for withUnretained where Element == Void


yoavschwartz:
### Name and description
I suggest adding an overload for `withUnretained` where the element type of the `Observable` is `Void` where the `Void` value will not be "forwarded" as its useless.
### Motivation for inclusion
I really enjoy the new operator, however, I believe at least 80% of the uses are to inject a safe strong reference of `self` to the subscribe block. When the `Observable` is of `Observable<Void>` which is very common in UI events, for example a `button.rx.tap`, we want to inject self to the `subscribe` but we really don't need the value. However, because the `Void` is sent along instead we have to explicitly use 
```
observableOfVoid.withUnretained(self).subscribe(onNext: { (strongSelf, _) in ... }
```
Instead I suggest making an overload that ignores void thus it would become
```
observableOfVoid..withUnretained(self).subscribe(onNext: { strongSelf in ... }
```
### Example of use

```
button.rx.tap.withUnretained(self)
.subscribe(onNext: { strongSelf in strongSelf.showButtonPressedAlert() }
```

### Implementation
I already have an implementation ready since I use it in my code, however since this is not a bug fix, seems like a discussion is due if this is important enough to be included.
```
extension ObservableType where E == Void {
    /**
     Provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object replacing the element emitted by the sequence in case its `Void`.
     In the case the provided object cannot be retained successfully, the seqeunce will complete.
     
     - parameter obj: The object to provide an unretained reference on.
     - parameter resultSelector: A function to transform the unretained referenced on `obj`.
     - returns: An observable sequence that contains the result of `resultSelector` being called with an unretained reference on `obj`.
     */
    public func withUnretained<T: AnyObject, Out>(_ obj: T,
                                                  resultSelector: @escaping (T) -> Out) -> Observable<Out> {
        let fun: (T, Self.E) -> Out = { (element, _) in resultSelector(element) }
        return withUnretained(obj, resultSelector: fun)
    }
    
    /**
     Provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object replacing the element emitted by the sequence in case its `Void`.
     In the case the provided object cannot be retained successfully, the seqeunce will complete.
     
     - parameter obj: The object to provide an unretained reference on.
     - returns: An observable sequence that contains an unretained reference on `obj`.
     */
[...]

PR #219, opened 2019-06-25
Add withUnretained for Element == Void

MortyMerr:
https://github.com/RxSwiftCommunity/RxSwiftExt/issues/178

### DIFF
--- a/Source/RxSwift/withUnretained.swift
+++ b/Source/RxSwift/withUnretained.swift
@@ -46,6 +46,32 @@ extension ObservableType {
     }
 }
 
+extension ObservableType where Element == Void {
+    /**
+     Provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object replacing the element emitted by the sequence in case its `Void`.
+     In the case the provided object cannot be retained successfully, the seqeunce will complete.
+
+     - parameter obj: The object to provide an unretained reference on.
+     - parameter resultSelector: A function to transform the unretained referenced on `obj`.
+     - returns: An observable sequence that contains the result of `resultSelector` being called with an unretained reference on `obj`.
+     */
+    public func withUnretained<T: AnyObject, Out>(_ obj: T, resultSelector: @escaping (T) -> Out)
+        -> Observable<Out> {
+            return withUnretained(obj) { object, _ in resultSelector(object) }
+    }
+
+    /**
+     Provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object replacing the element emitted by the sequence in case its `Void`.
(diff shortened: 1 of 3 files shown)


freak4pc:
@MortyMerr please resolve conflicts.

Also, can you explain what's wrong with the existing impl.? It doesn't work for Void because of the resultSelector ?

MortyMerr:
Hey @freak4pc sure. The purpose of the initial issue was to ignore Void element, only passing `self` parameter to result selector. I guess just for better syntax

Source/RxSwift/withUnretained.swift
```diff
@@ -46,6 +46,32 @@ extension ObservableType {
...
+     Provides an unretained, safe to use (i.e. not implicitly unwrapped), reference to an object replacing the element emitted by the sequence in case its `Void`.
+     In the case the provided object cannot be retained successfully, the seqeunce will complete.
+
+     - parameter obj: The object to provide an unretained reference on.
+     - parameter resultSelector: A function to transform the unretained referenced on `obj`.
+     - returns: An observable sequence that contains the result of `resultSelector` being called with an unretained reference on `obj`.
+     */
+    public func withUnretained<T: AnyObject, Out>(_ obj: T, resultSelector: @escaping (T) -> Out)
+        -> Observable<Out> {
+            return withUnretained(obj) { object, _ in resultSelector(object) }
```
freak4pc:
```suggestion
        return withUnretained(obj) { object, _ in resultSelector(object) }
```

MortyMerr:
We can change name to something like `unretained`
This repo has few operators which save just few symbols. But they are used so often, that it makes sense to use them. For instance, `mapTo(1)` vs `map { _ in 1 }`. This is just sugar which improves code readability.

`withUnretained(self).subscribe(onNext: { $0.doSomething })` looks better for me than `withUnretained(self).subscribe(onNext: { (sself, _) in sself.doSomething })`
But of course, it's really subjective.

> there's no specific subscribe / drive / emit for Void-emitting objects
Yes, but this repo extends the main repo functionality, so I don't think it's a valid argument to not add this sugar.

Shortly, I think it might be useful. However, I have no personal interest in this operator, just working on closing issues and fulfilling user requests. If it's not useful, feel free to close this PR & initial issue

freak4pc:
> Yes, but this repo extends the main repo functionality, so I don't think it's a valid argument to not add this sugar.

My point was that even Ext doesn't aim to add these things. I find it a bit strange and adds relatively little value. Yes, this repo extends things, but every piece of code means more maintenance and a larger code base, so it needs to be things that have a solid use case and are highly useful and intuitive. 

I think this isn't the case here. I'll be closing for now but let me know if you have any other thoughts / ideas around this.

PR #220, opened 2019-06-25
Fix swiftlint warnings

MortyMerr:


Source/RxSwift/and.swift
```diff
@@ -76,7 +76,7 @@ extension ObservableType where Element == Bool {
 						completed += 1
 						guard completed == collection.count else { return }
 						// if all emitted at least one `true`, emit true otherwise just complete
-						if emitted.reduce(true, { $0 && $1 }) {
+						if emitted.allSatisfy({ $0 }) {
```
freak4pc:
```suggestion
						if emitted.allSatisfy { $0 } {
```

MortyMerr: Thank you for your comments :)
But I am not sure that this code will compile successfully

freak4pc: Hey - I believe it will, why wouldn't it? It's a regular trailing closure

image

freak4pc: Ah apparently it doesn't work in the context of `if`s. TIL :)

PR #227, opened 2019-08-09
Deprecate filterMap

MortyMerr:
```swift
.compactMap { predicate($0) ?$0 : nil }
// prefered over
.filterMap { predicate($0)  ? .map($0) : nil }
```

### DIFF
--- a/Source/RxSwift/filterMap.swift
+++ b/Source/RxSwift/filterMap.swift
@@ -22,7 +22,7 @@ extension ObservableType {
         - returning `.ignore` will filter the value out of the returned Observable
         - returning `.map(newValue)` will propagate newValue through the returned Observable.
      */
-    @available(*, deprecated, renamed: "compactMap", message: "compactMap was added to the main RxSwift repository and should be used instead of filterMap")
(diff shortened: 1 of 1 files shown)


Source/RxSwift/filterMap.swift
```diff
@@ -22,6 +22,7 @@ extension ObservableType {
         - returning `.ignore` will filter the value out of the returned Observable
         - returning `.map(newValue)` will propagate newValue through the returned Observable.
      */
+    @available(*, deprecated, renamed: "compactMap", message: "compactMap was added to the main RxSwift repository and should be used instead of filterMap")
```
freak4pc:
```suggestion
    @available(*, deprecated, renamed: "compactMap", message: "Use compactMap instead")
```

I think this is enough for a deprecation message

MortyMerr:
I think, we can wait on that. But IMO .filterMap definitely should be deprecated  :)

freak4pc:
Definitely agreed about deprecation, I just think the timing is tricky. Lets set this aside for now:)

PR #229, opened 2019-08-22
Avoid using force-unwraps of Optionals during .unwrap()

Willamin:
#trivial

### DIFF
--- a/Source/RxSwift/unwrap.swift
+++ b/Source/RxSwift/unwrap.swift
@@ -18,6 +18,6 @@ extension ObservableType {
      */
 
     public func unwrap<T>() -> Observable<T> where Element == T? {
-        return self.filter { $0 != nil }.map { $0! }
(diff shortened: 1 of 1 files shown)


MortyMerr:
Hi @Willamin thanks for contribution.
`.filter { $0 != nil }.map { $0!` can be replaced with `.compactMap`

freak4pc:
As Anton mentioned, `.unwrap` should not be used at this point since we we have `compactMap`. `unwrap` will be deprecated in a coming version of RxSwiftExt, but I think we should also make it use `compactMap` internally just so we're clear on it.

freak4pc:
Do any of you want to make this use `compactMap` for now ?

MortyMerr:
Sure, here https://github.com/RxSwiftCommunity/RxSwiftExt/pull/234

Issue #224, opened 2019-07-19
Documentation for `filterMap` is wrong


kroussevrb:
Documentation for `filterMap` is wrong. The comment in the code snippet says "keeps odd numbers". The text below the code snippet says "keeps only even numbers"

freak4pc:
Thanks :) will gladly accept a PR

PR #231, opened 2019-08-26
Fix filterMap doc #trivial

MortyMerr:
https://github.com/RxSwiftCommunity/RxSwiftExt/issues/224


freak4pc:
As long as we're here, want to change the underlying impl. to use compactMap ? As long as we're not breaking the public API I'm OK with it.

freak4pc:
We won't use deprecation annotations right now anyways - we'll leave the method as is and just use `compactMap` below the hood IMO (makes more sense for now). When we switch a major version we can deprecate.

PR #236, opened 2019-09-13
Add SharedSequence weak subscriptions

mpsnp:
This pull request adds both Signal & Driver weak subscriptions.

### DIFF
--- a/RxSwiftExt.xcodeproj/project.pbxproj
+++ b/RxSwiftExt.xcodeproj/project.pbxproj
@@ -69,7 +69,7 @@
 		538607EB1E6F36A9000361DE /* RetryWithBehaviorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 538607C81E6F367A000361DE /* RetryWithBehaviorTests.swift */; };
 		538607EC1E6F36A9000361DE /* UnwrapTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 538607C91E6F367A000361DE /* UnwrapTests.swift */; };
 		538607ED1E6F36A9000361DE /* WeakTarget.swift in Sources */ = {isa = PBXBuildFile; fileRef = 538607CA1E6F367A000361DE /* WeakTarget.swift */; };
-		538607EE1E6F36A9000361DE /* WeakTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 538607CB1E6F367A000361DE /* WeakTests.swift */; };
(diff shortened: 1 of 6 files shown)


freak4pc:
Is this really different then `withUnretained` + subscribe ?

MortyMerr:
Same question here 

mpsnp:
Ok, here is my explanation:

1. `withUnretained` + `subscribe` only works for `Observable`, so shared sequences should be converted to `Observable` first, loosing their original types, which is not good i guess
2. There are already extensions for `Observable` which do exactly the same as in this PR.
3. Using curried style is preferred in my opinion, because it does not require to wrap a call of "unretained" object function to closure or uncurry its method.  
    Just compare usage:
    ```swift
    state // Usually Driver
        .asObservable()
        .withUnretained(self)
        .subscribe(onNext: { controller, state in 
            controller.render(state: state)
        })
        .disposed(by: disposeBag)
    ```
    with
    ```swift
    state
        .drive(weak: self, onNext: Controller.render)
        .disposed(by: disposeBag)
    ```

mpsnp:
@freak4pc @MortyMerr any feedback? Actually this is not "new" operator, it already exists here but only for `ObservableType`. I just wanted to add it to `SharedSequence`. IMHO `withUnretained ` also should be added to Traits, what do you think?

MortyMerr:
Please also note this PR with corresponding issue https://github.com/RxSwiftCommunity/RxSwiftExt/pull/219
For me `subscribe(weak: self, onNext: ...)` and `withUnretanied(self).subscribe(onNext:)` purpose looks absolutely the same.

I think that these operators should be merged into one (for all traits). One operator for weakly referencing an object in subscription block is more than enough.

I prefer `withUnretained`, cause it is more generic. It can be used not only with subscription, but with `map`, `do` etc.

`subscribe(weak:` is too specific case, I think it's senseless to have own wrapper for all RxSwift operators (`do(weak:`, `map(weak:`, `filter(weak:`)

What do you think?

freak4pc:
I'm siding with Anton here — I'm seeing very little value adding a bunch of **do something with weak object** overloads when we have `withUnretained` that solves all of these cases. Yes, there is an existing `subscribe(onNext:)` but it precedes `withUnretained` by a years, probably :) There's no reason to deprecate it since it's already in the library, but there's no reason to add more of it with a better solution available.

Currying-style doesn't seem like a good enough reason (to me) to add a bunch of these overloads.

MortyMerr:
What we actually can do is to add withUnretained for all traits

PR #239, opened 2019-10-01
Reduce one character generics usage #trivial

MortyMerr:
It's better for readability to avoid one character generics naming. RxSwift already replaced all generics to meaningful title, it's RxSwiftExt turn

P.S. lol, i found that project uses tabs as indents. I am not sure, should i indent new code with tabs or spaces

### DIFF
--- a/Source/RxSwift/cascade.swift
+++ b/Source/RxSwift/cascade.swift
@@ -20,7 +20,7 @@ extension Observable where Element: ObservableType {
 	- parameter observables: a sequence of observables which will all be immediately subscribed to
 	- returns: An observable sequence that contains elements from the latest observable sequence that emitted elements
 	*/
-	public static func cascade<S: Sequence>(_ observables: S) -> Observable<Element.Element> where S.Element == Element {
+    public static func cascade<Sequebce: Swift.Sequence>(_ observables: Sequebce) -> Observable<Element.Element> where Sequebce.Element == Element {
         let flow = Array(observables)
         if flow.isEmpty {
             return Observable<Element.Element>.empty()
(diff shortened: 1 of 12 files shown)


Source/RxSwift/cascade.swift
```diff
@@ -20,7 +20,7 @@ extension Observable where Element: ObservableType {
 	- parameter observables: a sequence of observables which will all be immediately subscribed to
 	- returns: An observable sequence that contains elements from the latest observable sequence that emitted elements
 	*/
-	public static func cascade<S: Sequence>(_ observables: S) -> Observable<Element.Element> where S.Element == Element {
+    public static func cascade<Sequebce: Swift.Sequence>(_ observables: Sequebce) -> Observable<Element.Element> where Sequebce.Element == Element {
```
jdisho: nit: typo in `Sequebce`.

Source/RxSwift/zipWith.swift
```diff
@@ -21,7 +21,7 @@ extension ObservableConvertibleType {
 
      - Returns: An observable sequence containing the result of combining elements of the sources using the specified result selector function.
      */
-    public func zip<O2, ResultType>(with second: O2, resultSelector: @escaping (Element, O2.Element) throws -> ResultType) -> Observable<ResultType> where O2: ObservableConvertibleType {
+    public func zip<Observable2: ObservableConvertibleType, ResultType>(with second: Observable2, resultSelector: @escaping (Element, Observable2.Element) throws -> ResultType) -> Observable<ResultType> {
```
jdisho: Can't we rename `Observable2` to `Observable`?

freak4pc: I'd just call it `Other` — that's how we do it at some places inside RxSwift

Source/RxSwift/cascade.swift
MortyMerr: Fixed, thanks

Source/RxSwift/zipWith.swift
MortyMerr: Fixed, thanks

Issue #242, opened 2019-10-04
Add withUnretained for all traits


MortyMerr:
Currently, `withUnretained` defined only for Observable. Wouldn't it be useful to add it for all traits?

PR #243, opened 2019-10-04
Add withUnretained to SharedSequence

freak4pc:
Based on the request in #242. IRT other traits - there's no real value to add it to Completable... we could add to Maybe and Single in a separate PR.

This PR has a single failing test, I can't really put my finger on what's going on there (the same test passes for Signal but fails for Driver). 

If anyone can take a poke at it, that'd be great :) 

Thanks !


Source/RxCocoa/withUnretained+RxCocoa.swift
```diff
@@ -0,0 +1,43 @@
...
+     In the case the provided object cannot be retained successfully, the seqeunce will with an error.
+
+     - parameter obj: The object to provide an unretained reference on.
+     - parameter resultSelector: A function to combine the unretained referenced on `obj` and the value of the observable sequence.
+     - returns: A shared sequence that contains the result of `resultSelector` being called with an unretained reference on `obj` and the values of the original sequence.
+     */
+    public func withUnretained<Object: AnyObject, Out>(_ obj: Object,
+                                                       resultSelector: @escaping ((Object, Element)) -> Out)
+        -> SharedSequence<SharingStrategy, Out> {
+        asObservable()
```
freak4pc: I'm casting back and forth to be able to throw a sequence-terminating error. The original implementation used a flatMap but that would be tricky given a completion doesn't affect the upstream.

Source/RxSwift/withUnretained.swift
```diff
@@ -42,10 +42,10 @@ extension ObservableType {
      - returns: An observable sequence of tuples that contains both an unretained reference on `obj` and the values of the original sequence.
      */
     public func withUnretained<Object: AnyObject>(_ obj: Object) -> Observable<(Object, Element)> {
-        return withUnretained(obj) { ($0, $1) }
```
MortyMerr: Package.swift has `.v4_2` version. Is it ok to migrate to Swift 5.1 syntax? I think it's better to perform this migration in separate PR for whole repo.

Tests/RxCocoa/withUnretainedTests+RxCocoa.swift
```diff
@@ -0,0 +1,218 @@
...
+//  Copyright © 2019 RxSwift Community. All rights reserved.
+//
+
+import XCTest
+import RxCocoa
+import RxSwift
+import RxTest
+
+class WithUnretainedCocoaTests: XCTestCase {
+    fileprivate var testClass: TestClass!
```
MortyMerr: nit: all properties could be private

freak4pc: Good point!

Source/RxSwift/withUnretained.swift
freak4pc: Uhm yeah, we'll just do the migration in a separate PR.

Once RxSwift 6 is out, it will only support Xcode 11 & Swift 5.1. So we'll release RxSwiftExt 6 accordingly.

getogrand:
I found a bug on this implementation.

In below scenario, the `viewModel.estimatedShippingDate` is not disposed, so that the viewController and viewModel were not removed in memory.
I don't really understand why this causes retain cycle nevertheless the `self` is not referred in the closure body.

The retain cycle is not caused if I change `.withUnretained(self)` with normal `[weak self]` capture list.

```swift
enum YGShippingType: String, Codable, Comparable, CustomStringConvertible {
  // ...
}

class SomeViewModel {
  private let _estimatedShippingDate = PublishRelay<(Date, YGShippingType)?>()

  // ...

  var estimatedShippingDate: Driver<(Date, YGShippingType)?> {
    _estimatedShippingDate.asDriver(onErrorJustReturn: nil)
      .distinctUntilChanged({
        $0?.0 == $1?.0 && $0?.1 == $1?.1
      })
  }
}

class SomeViewController: UIViewController {
  private let disposeBag = DisposeBag()
  var viewModel: ViewModel! // Injected by swinject

  // ...

  override func viewDidLoad() {
    super.viewDidLoad()
    bindCartTableHeaderView()
  }

  private func bindCartTableHeaderView() {
    viewModel.estimatedShippingDate // Driver<(Date, YGShippingType)?>
      .map({ $0 == nil })
      .distinctUntilChanged()
      .withUnretained(self)
      .drive(onNext: { (self, hide) in
        // Note that we did not refer `self`
        print("blabla")
      })
      .disposed(by: disposeBag)
  }
}
```

PR #245, opened 2019-11-08
Enable watchos for carthage

dchohfi:
Adding WatchOS target to the project that enables Carthage to build it.


dchohfi:
@freak4pc does this makes sense?

PR #248, opened 2020-05-26
Link to RxCocoa on macOS & tvOS targets

Lucconouche:
Hello RxSwiftExt maintainers!

The main app I work on uses RxSwiftExt and does not use any package managers to integrate it. RxSwiftExt is integrated in a codebase containing both macOS and iOS code. We ran into an issue where our Mac target would no longer compiled saying RxCocoa was missing when compiling RxSwiftExt. This issue was non-existant when using RxSwiftExt with our iOS target.

While looking at the Package.swift & the RxSwiftExt.podspec files, the dependency to RxSwift is set correctly. When you integrate the Xcode Project to compile RxSwiftExt yourself, only the iOS target was linking correctly to its dependencies.

Moreover, I removed linking to RxRelay since no target was using it.

### DIFF
--- a/RxSwiftExt.xcodeproj/project.pbxproj
+++ b/RxSwiftExt.xcodeproj/project.pbxproj
@@ -22,6 +22,8 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


Lucconouche:
Thanks @freak4pc!

PR #253, opened 2020-09-27
distinct:  use filter instead of flatMap #trivial

hoc081098:
I think we can replace `flatMap` by `compactMap` for better performance in `distinct.swift`, without adding anything to the Changelog.

### DIFF
--- a/Source/RxSwift/distinct.swift
+++ b/Source/RxSwift/distinct.swift
@@ -19,13 +19,10 @@ extension Observable {
      */
     public func distinct(_ predicate: @escaping (Element) throws -> Bool) -> Observable<Element> {
         var cache = [Element]()
-        return compactMap { element -> Element? in
-            if try cache.contains(where: predicate) {
-                return nil
-            } else {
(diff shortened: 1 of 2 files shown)


Source/RxSwift/distinct.swift
```diff
@@ -19,12 +19,12 @@ extension Observable {
...
         var cache = [Element]()
-        return flatMap { element -> Observable<Element> in
+        return compactMap { element -> Element? in
             if try cache.contains(where: predicate) {
-                return Observable<Element>.empty()
+                return nil
             } else {
                 cache.append(element)
-                return Observable<Element>.just(element)
+                return element
```
freak4pc, line 27: Wondering if it's worth using a set and filter similarly to what we've done in CombineExt. WDYT?
https://github.com/CombineCommunity/CombineExt/blob/master/Sources/Operators/RemoveAllDuplicates.swift#L22-L24

hoc081098: I agree with you

PR #267, opened 2021-05-14
Added RxSwift submodule dependency

philippec:
There is a reference to Rx.xcodeproj in the project, and without this it does not build if you don't use Carthage.

It's essentially the same fix as https://github.com/RxSwiftCommunity/RxDataSources/commit/98ae6d937f37766187f97cd066076fa7127cff01


freak4pc (approved):
Looks good, thanks!

PR #270, opened 2022-05-11
Update podspec version to match release

lordzsolt:
After 6.1.0 release was created, the .podspec file was not update, so you are unable to use 6.1.0 in your Podfile.

### DIFF
--- a/RxSwiftExt.podspec
+++ b/RxSwiftExt.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "RxSwiftExt"
-  s.version      = "6.0.1"
+  s.version      = "6.1.0"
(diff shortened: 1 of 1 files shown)


freak4pc (approved):

RxSwiftExt.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "RxSwiftExt"
-  s.version      = "6.0.1"
+  s.version      = "6.1.0"
```
freak4pc, line 3: I think we'll have to make this 6.1.1 - 
The 6.1.0 release is already locked to a commit without the updated pod spec, so `pod lib lint` won't let me upload it to the Specs repo

freak4pc (changes requested):

PR #275, opened 2023-09-06
Added Observable+Alias.swift to project file with correct targets. #trivial

adamsousa:
Observable-Alias.swift was missing targets and references in the project file. Added in this PR

#trivial


freak4pc:
Thanks for the fix, but I'm not sure adding it to the project is relevant at all to fetching it via SPM or even CocoaPods. It's only for the Sample Project itself, unless you're referring to Carthage?

adamsousa:
We are using Carthage, I wasn't able to rebuild this framework until this file was added to the project file.

marcbaldwin:
I was also unable to build this framework using Carthage. Thanks for the fix, now that Xcode15 has been released, it'd be great to get this released too 🙏

freak4pc:
Thanks!

